PR #45, opened 2017-03-03
Remove mkdir -p where neccessary

jeanlaurent:
- `mkdir -p` with only a subdir is not useful
- it creates a `-p` directory in windows cmd

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ deps:
 	go get github.com/golang/lint/golint
 
 osxkeychain:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -8,7 +8,7 @@ deps:
 	go get github.com/golang/lint/golint
 
 osxkeychain:
-	mkdir -p bin
+	mkdir bin
```
vdemeester: It could even be a target on its own 👼

Issue #48, opened 2017-03-09
secretservice tests fail


n4ss:
**Expected behavior**:

All tests ran by `make test` for each platform return `ok`.

**Experienced behavior**:

Every once in a while (totally random), I obtain the following output on linux (thus with `secretservice`):
``` shell
$ make test
# tests all packages except vendor
go test -v `go list ./... | grep -v /vendor/`
=== RUN   TestStore
--- PASS: TestStore (0.00s)
=== RUN   TestGet
--- PASS: TestGet (0.00s)
=== RUN   TestErase
--- PASS: TestErase (0.00s)
=== RUN   TestList
--- PASS: TestList (0.00s)
PASS
ok  	github.com/docker/docker-credential-helpers/client	0.002s
=== RUN   TestStore
--- PASS: TestStore (0.00s)
=== RUN   TestGet
--- PASS: TestGet (0.00s)
=== RUN   TestErase
--- PASS: TestErase (0.00s)
=== RUN   TestList
--- PASS: TestList (0.00s)
PASS
ok  	github.com/docker/docker-credential-helpers/credentials	0.002s
=== RUN   TestSecretServiceHelper
--- FAIL: TestSecretServiceHelper (0.15s)
	secretservice_linux_test.go:45: <nil>
=== RUN   TestMissingCredentials
--- PASS: TestMissingCredentials (0.00s)
FAIL
FAIL	github.com/docker/docker-credential-helpers/secretservice	0.154s
?   	github.com/docker/docker-credential-helpers/secretservice/cmd	[no test files]
Makefile:28: recipe for target 'test' failed
make: *** [test] Error 1
```
This indicates an error for the following test:
``` go
if newauths, err := helper.List(); (len(newauths) - len(auths)) != 1 {
	t.Fatal(err)
}
```

When printing the number of new and old credentials: `len(newauths)` and `len(auths)`, I see the same number twice. This shouldn't happen if the lists obtained from `helper.List()` are only docker credentials. Since it's not currently the case and that the lists we obtain are all entries from the default secret service, we can get a race condition here.

Also the check in `secretservice/secretservice_linux_test.go` at line 40 doesn't make sense to me:
``` go
if err != nil || len(auths) == 0 {
```
We add credentials, then delete the entry and expect the number of credentials to be non-zero. We should expect a 0 value.

In issue https://github.com/docker/docker-credential-helpers/issues/46, I propose to filter out non-Docker credentials from the results obtained by `Helper.List()`.

[...]

Issue #47, opened 2017-03-08
docker-credential-osxkeychain asking for access to keychains and passwords it has no need for


mightyroser:
When I perform a docker build of a very simple docker file, docker-credential-osxkeychain prompts
me for access to my bitbucket.org password stored in my keychain.  There is absolutely no reason why
docker would need access to this password.

xxx@yyy ~/docker_test [0]$ docker --version
Docker version 17.03.0-ce, build 60ccb22
xxx@yyy ~/docker_test [0]$ vi Dockerfile
xxx@yyy ~/docker_test [0]$ cat Dockerfile 
FROM python:3.5-alpine
xxx@yyy ~/docker_test [0]$ docker build -t docker_build_creds_test .
Sending build context to Docker daemon 2.048 kB
Step 1/1 : FROM python:3.5-alpine
 ---> e3c9c70fb387
Successfully built e3c9c70fb387
xxx@yyy ~/docker_test [0]$ 

<img width="437" alt="screen shot 2017-03-08 at 10 26 38 am" src="https://cloud.githubusercontent.com/assets/25228271/23710219/f88520c2-03e9-11e7-9631-842fcd3189ce.png">

vdemeester:
@mightyroser thanks for the report, this is tracked [here](https://github.com/docker/docker/issues/31517) and [here](https://github.com/docker/docker-credential-helpers/issues/46) 👼

vdemeester:
This should be fixed with #49 merged and `v0.5.0` release. Will be closed once docker4mac update is there 👼

n4ss:
This is fixed as of v0.5.0

Issue #46, opened 2017-03-08
Add a label for Docker credentials and filter credentials lookup


n4ss:
This is a follow-up on these issues:

- https://github.com/docker/for-mac/issues/1371

- https://github.com/docker/docker/issues/31517

We'd want to tag the credentials we store, that way we would be able to filter out any non-Docker credentials when looking for the proper item.

I propose to write a patch with a "Docker Credentials" label for Docker credentials with:

- osxkeychain: `kSecAttrLabel`
- wincred: `GenericCredential.Attributes`
- secretservice: `SecretItem attributes`

We would also be able later on to provide "sub-labelling" through attributes (and kSecAttrDescription on macOS) to differentiate the credentials for different Docker services we want to log into (for now we log in Docker registries but I'm just thinking about potential future constraints).

n4ss:
Fix is being rolled out on different platforms.

Issue #32, opened 2016-12-13
Unable to push to Registry V2 with docker-compose using credstore on linux


hamid-elaosta:
When using the secretservice setting for the credsStore (the default), I am unable to push into Registry V2 using docker-compose 1.8.0 (always returns "no basic auth credentials"). After removing the credsStore entry from the configuration and using inline storage, everything works fine.

PR #49, opened 2017-03-10
Label credentials on each platform' creds store and fix secretservice behavior

n4ss:
Logic behind this PR's purpose and implementation are here:  https://github.com/docker/docker-credential-helpers/issues/46.

This PR is solving the following issues:

- https://github.com/docker/docker-credential-helpers/issues/27
- https://github.com/docker/docker-credential-helpers/issues/46
- https://github.com/docker/docker-credential-helpers/issues/14
- https://github.com/docker/docker-credential-helpers/issues/47

and probably solving the following issue:

- https://github.com/docker/docker-credential-helpers/issues/32

And also solves one of the two reasons behind this issue:

- https://github.com/docker/docker-credential-helpers/issues/48

The `"Docker Credentials"` label is now added to credentials on `osxkeychain` (macOS), `wincred` (Windows) and `secretservice` (Linux) to filter out non-Docker credentials when listing those.

`secretservice` was broken in several spots, this is also fixed in this PR. The tests for this platform are fixed in the PR associated to this issue: https://github.com/docker/docker-credential-helpers/issues/48.

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -106,8 +106,8 @@ func ExampleStore() {
 
 func TestStore(t *testing.T) {
 	valid := []credentials.Credentials{
-		{credentials.CredsLabel, validServerAddress, "foo", "bar"},
-		{credentials.CredsLabel, validServerAddress2, "<token>", "abcd1234"},
+		{validServerAddress, "foo", "bar"},
+		{validServerAddress2, "<token>", "abcd1234"},
 	}
 
 	for _, v := range valid {
@@ -117,7 +117,7 @@ func TestStore(t *testing.T) {
 	}
 
 	invalid := []credentials.Credentials{
-		{credentials.CredsLabel, invalidServerAddress, "foo", "bar"},
+		{invalidServerAddress, "foo", "bar"},
 	}
 
 	for _, v := range invalid {
@@ -140,8 +140,8 @@ func ExampleGet() {
 
 func TestGet(t *testing.T) {
 	valid := []credentials.Credentials{
-		{credentials.CredsLabel, validServerAddress, "foo", "bar"},
-		{credentials.CredsLabel, validServerAddress2, "<token>", "abcd1234"},
+		{validServerAddress, "foo", "bar"},
+		{validServerAddress2, "<token>", "abcd1234"},
 	}
 
 	for _, v := range valid {
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -106,8 +106,8 @@ func ExampleStore() {
 
 func TestStore(t *testing.T) {
 	valid := []credentials.Credentials{
-		{validServerAddress, "foo", "bar"},
-		{validServerAddress2, "<token>", "abcd1234"},
+		{credentials.CredsLabel, validServerAddress, "foo", "bar"},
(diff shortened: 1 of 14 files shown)


simonferquel:
Hmm, what about existing users already logged in with the previous behavior ? I suppose they will appear as logged in and can do pull/pushes but will fail to build some dockerfiles?

vdemeester:
@simonferquel yep, they would need to logout/login again. Not so much we can do about it in this project though — we would over-complexify something for not too much of a benefit. Somehow it should be handled downstream (big warning in release note, or force users to logout/login, …)

jeanlaurent:
This PR allow us to workaround ( and workaround only) the issue at a quicker pace since this does not involve a docker/docker release, but I feel like the long term solution must be fixed in docker/docker ( https://github.com/docker/docker/issues/31517 )

simonferquel:
Hmm, i am not fond of the idea. Could we make a change in docker cli to pass the list of authenticated servers from the client json file?

ebriney:
We can a FAQ and message in the update window (changelogs) for users that are already logged

endophage:
Would be great to expose the CredsLabel as an argument at the higher level interfaces. Notary was going to make use of this repo but we'd probably want to use a different label.

n4ss:
@jeanlaurent I agree that the long term solution must be fixed in docker/docker although namespacing and tagging should be present too in credentials stores that allow it.

@ebriney sounds good to me.

@endophage when you say "higher level interfaces", are you talking about the `Helper` interface from `credentials/helper.go`? I didn't want to break the tons of helpers that people have built on top of it (AWSLabs, GCP etc..).

endophage:
@n4ss that's fair, I wasn't sure what the scope of dependents was. Would still be great though to provide some way for a consumer of this library to be able to provide an alternate label. Maybe make it a `var` not a `const` and provide a setter?

n4ss:
@endophage Sounds good, will fix.

vdemeester:
@endophage @n4ss or maybe functionnal/variadic arguments with a config struct 👼

n4ss:
@vdemeester won't golang complain with a type mismatch for `List(label ...string) != List()` ?

vdemeester:
@n4ss yeah I though of that. And you are right, it won't satisfy existing interface so… 😅 
Let's ping @samuelkarp @KingEmet to see if it could make sense (and if we do this so you could prepare yourself)

n4ss:
@vdemeester I think labels shouldn't be a part of `Helper` as it is just optional to tag credentials, your own creds store can of course have other mechanisms to identify credentials groups. But I think it's fair to expose a default credentials label string in the `credentials` package.

KingEmet:
Agreed with @n4ss, it should probably be left as an implementation detail in the cred helper.

n4ss:
@vdemeester @simonferquel @ebriney @jeanlaurent do we agree on the following solution?
> FAQ and message in the update window (changelogs) for users that are already logged

@KingEmet @endophage does the PR look good to you?

simonferquel:
We'll try to find a way to log out users automatically by reading the client conf file (at least if we find one at its default location)

PR #50, opened 2017-03-10
Fix secretservice tests and expected behaviors

n4ss:
The logic behind this PR's purpose and implementation can be found here: https://github.com/docker/docker-credential-helpers/issues/48

This PR needs the following PR to work properly: https://github.com/docker/docker-credential-helpers/pull/49 which fixes the broken `secureservice` implementation.

### DIFF
--- a/secretservice/secretservice_linux_test.go
+++ b/secretservice/secretservice_linux_test.go
@@ -26,16 +26,13 @@ func TestSecretServiceHelper(t *testing.T) {
 
 	// If any docker credentials with the tests values we are providing, we
 	// remove them as they probably come from a previous failed test
-	if len(old_auths) >= 1 {
-		for k, v := range old_auths {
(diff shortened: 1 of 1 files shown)


secretservice/secretservice_linux_test.go
```diff
@@ -16,10 +17,39 @@ func TestSecretServiceHelper(t *testing.T) {
...
+
+	// Check how many docker credentials we have when starting the test
+	old_auths, err := helper.List()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// If any docker credentials with the tests values we are providing, we
+	// remove them as they probably come from a previous failed test
+	if len(old_auths) >= 1 {
```
vdemeester: Not sure if it's required here. The for loop just won't do a thing if the length of `old_auths` is `0` 👼

n4ss: Will fix, thanks.

Issue #14, opened 2016-04-14
OSX Keychain storage behavior in Keychain Access is strange


seanlutner:
I was testing out the new credential helpers this morning, specifically the osxkeychain method. I went through the example of logging in to hub.docker.com, and I saw that the Keychain Access entry was created. I also have an internal registry, which I was able to authenticate to and see an entry created, but only the username was stored. The login URL/host was not stored in the Where field, I only see http://, which is odd because my repo is SSL protected. See attached screenshots. I'm not sure that it matters operationally, but it would be nice if all registry authentications were stored properly. I might have the same username against multiple internal registries.

To login to the internal registry, I did `docker login my.registryhost.com:5000`

![Uploading registry1.png…]()
![Uploading registry2.png…]()

calavera:
it looks like the screenshots were not attached, but I agree that it should work with any registry. I wonder if the URL is not properly populated in the docker side :pensive:

seanlutner:
Whoops, I didn't realize they hadn't uploaded. Trying again. I have a related question, and I'll open another issue if required, does the linux helper require a full X environment based on it's dependency on gnome-keyring? I'm looking to use this from headless server systems and don't want to have to install full X everywhere.

![registry1](https://cloud.githubusercontent.com/assets/9679816/14560728/6d79b792-02df-11e6-9d85-f40e2b777bf6.png)
![registry2](https://cloud.githubusercontent.com/assets/9679816/14560730/7227687a-02df-11e6-8654-a0ad1ee21f92.png)

dematic-denis-antonioli:
An other weird behavior I'm using the google container repository, logging in with
`docker login -u oauth2accesstoken -p "$(gcloud auth print-access-token)" us.gcr.io`
and the 'where' field is set to `http:///us.gcr.io`

I expect it to be `https://us.gcr.io`

<img width="259" alt="oauth2accesstoken" src="https://cloud.githubusercontent.com/assets/20247179/21191380/39548e78-c225-11e6-80c9-b51cd20bed63.png">

KingEmet:
@dematic-denis-antonioli one workaround for that is to log in to 'https://us.gcr.io' explicitly:
`docker login -u oauth2accesstoken -p "$(gcloud auth print-access-token)" https://us.gcr.io`

fajpunk:
aws-cli/1.11.61 Python/3.5.0 Darwin/15.6.0 botocore/1.5.24
[...]

PR #55, opened 2017-03-15
Force serverURL to have https protocol

fajpunk:
See https://github.com/docker/docker-credential-helpers/issues/14#issuecomment-286532314

Docker login strips the protocol from serverURL, and docker pull prepends an `http` protocol to serverURL, resulting in:

1. creds from docker login not being stored correctly due to `net/url.Parse` not parsing protocol-less URLs correctly
2. Creds stored with a protocol-less URL not being able to be retrieved with a URL that has a protocol

I'm not sure the correct thing to do here, but since this url parsing is just done to store and retrieve credentials from the OSX keychain, it seems like it's OK to just normalize all `serverURL`s to have the `https` protocol.  This way, `net/url.Parse` will always parse them as expected, and they'll match among any different way that the docker command could pass them.

### DIFF
--- a/osxkeychain/osxkeychain_darwin.go
+++ b/osxkeychain/osxkeychain_darwin.go
@@ -11,6 +11,7 @@ import "C"
 import (
 	"errors"
 	"net/url"
+	"regexp"
 	"strconv"
 	"strings"
 	"unsafe"
@@ -134,8 +135,17 @@ func (h Osxkeychain) List() (map[string]string, error) {
 	return resp, nil
 }
 
+// alwaysHTTPS takes a url and makes sure it always has the https protocol
+func alwaysHTTPS(url string) string {
+	re := regexp.MustCompile(`^\w+://`)
+	stripped := re.ReplaceAllLiteralString(url, "")
+	return ("https://" + stripped)
+}
+
 func splitServer(serverURL string) (*C.struct_Server, error) {
-	u, err := url.Parse(serverURL)
+	httpsServerURL := alwaysHTTPS(serverURL)
+
+	u, err := url.Parse(httpsServerURL)
 	if err != nil {
 		return nil, err
 	}
--- a/osxkeychain/osxkeychain_darwin.go
+++ b/osxkeychain/osxkeychain_darwin.go
@@ -13,7 +13,6 @@ import (
 	"net/url"
 	"regexp"
 	"strconv"
-	"strings"
 	"unsafe"
 
 	"github.com/docker/docker-credential-helpers/credentials"
@@ -135,40 +134,47 @@ func (h Osxkeychain) List() (map[string]string, error) {
 	return resp, nil
 }
 
-// alwaysHTTPS takes a url and makes sure it always has the https protocol
-func alwaysHTTPS(url string) string {
-	re := regexp.MustCompile(`^\w+://`)
-	stripped := re.ReplaceAllLiteralString(url, "")
-	return ("https://" + stripped)
-}
-
+// splitServer() creates a proper server structure for OSX Keychain API
+// It normalizes if needed the format of the URL. Of no protocol is
+// provided, HTTPS will be used by default.
 func splitServer(serverURL string) (*C.struct_Server, error) {
-	httpsServerURL := alwaysHTTPS(serverURL)
+	// Check if we have a scheme in the URL as of RFC 3986, section 3.1
+	// and prepend '//' to normalize to a valid URL format
+	if !regexp.MustCompile(`^([a-zA-Z][-+.a-zA-Z0-9]+:)?//`).MatchString(serverURL) {
+		serverURL = "//" + serverURL
+	}
 
-	u, err := url.Parse(httpsServerURL)
+	u, err := url.Parse(serverURL)
 	if err != nil {
 		return nil, err
 	}
 
-	hostAndPort := strings.Split(u.Host, ":")
-	host := hostAndPort[0]
+	// We only support HTTPS and HTTP for registries
+	if u.Scheme != "" && u.Scheme != "https" && u.Scheme != "http" {
+		return nil, errors.New("unsupported scheme: " + u.Scheme)
+	}
+
+	if u.Hostname() == "" {
+		return nil, errors.New("no hostname in URL")
+	}
+
+	// If no protocol is specified, we use HTTPS
+	proto := C.kSecProtocolTypeHTTPS
+	if u.Scheme == "http" {
+		proto = C.kSecProtocolTypeHTTP
+	}
+
 	var port int
-	if len(hostAndPort) == 2 {
-		p, err := strconv.Atoi(hostAndPort[1])
+	if u.Port() != "" {
+		port, err = strconv.Atoi(u.Port())
 		if err != nil {
 			return nil, err
 		}
-		port = p
-	}
-
-	proto := C.kSecProtocolTypeHTTPS
-	if u.Scheme != "https" {
-		proto = C.kSecProtocolTypeHTTP
 	}
 
 	return &C.struct_Server{
 		proto: C.SecProtocolType(proto),
-		host:  C.CString(host),
+		host:  C.CString(u.Host),
 		port:  C.uint(port),
 		path:  C.CString(u.Path),
 	}, nil
--- a/osxkeychain/osxkeychain_darwin_test.go
+++ b/osxkeychain/osxkeychain_darwin_test.go
@@ -54,6 +54,68 @@ func TestOSXKeychainHelper(t *testing.T) {
 	}
 }
 
+func TestOSXKeychainHelperIgnoresProtocol(t *testing.T) {
+	loginServerURL := "foobar.docker.io:2376"
+	pullServerURL1 := "http://foobar.docker.io:2376"
+	pullServerURL2 := "https://foobar.docker.io:2376"
+	pullServerURL3 := "ftp://foobar.docker.io:2376"
+
+	creds := &credentials.Credentials{
+		ServerURL: loginServerURL,
+		Username:  "foobar",
+		Secret:    "foobarbaz",
+	}
+	helper := Osxkeychain{}
+	defer helper.Delete(creds.ServerURL)
+	if err := helper.Add(creds); err != nil {
+		t.Fatal(err)
+	}
+
+	username1, secret1, err := helper.Get(pullServerURL1)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if username1 != "foobar" {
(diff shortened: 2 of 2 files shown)


fajpunk:
Note that merging this will likely make it so that existing creds in the osx keychain will not be found; folks will either have to manually modify the existing keychain entries, or get new keys with this updated binary.

Maybe there's a better way to solve this than the one in this PR?

fajpunk:
Do I need to do anything with this PR given the Moby switch?

vdemeester:
@fajpunk no, we just need to take a little bit of time to review it 😓
/cc @n4ss too 👼

n4ss:
Sorry for the late review T_T
Can you try the tests again once fixed with the notes from my review?

osxkeychain/osxkeychain_darwin.go
```diff
@@ -134,8 +134,21 @@ func (h Osxkeychain) List() (map[string]string, error) {
 	return resp, nil
 }
 
+// alwaysHTTPS takes a url and makes sure it always has the https protocol
+func alwaysHTTPS(url string) string {
```
n4ss: Actually it would be sort of a pain, we should verify that `url.Parse` rejects these right after.
A test for that would probably be better, just checking that `ftp://foobar:1234` fails

fajpunk: I modified this to strip any protocol (regex: `^.*//`) with https.  Does that work?

osxkeychain/osxkeychain_darwin.go
```diff
@@ -134,8 +134,21 @@ func (h Osxkeychain) List() (map[string]string, error) {
...
+	stripped := url
+	if strings.HasPrefix(url, "http://") {
```
fajpunk: N/A if the regex solution is good

osxkeychain/osxkeychain_darwin_test.go
```diff
@@ -54,6 +54,57 @@ func TestOSXKeychainHelper(t *testing.T) {
...
+	}
+	if secret != "foobarbaz" {
+		t.Fatalf("Error: expected secret %s, got secret %s", creds.Secret, secret)
+	}
+
+	username2, secret2, err := helper.Get(pullServerURL2)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if username != "foobar" {
```
fajpunk: D'oh! Fixed

osxkeychain/osxkeychain_darwin_test.go
```diff
@@ -54,6 +54,57 @@ func TestOSXKeychainHelper(t *testing.T) {
...
+		t.Fatalf("Error: expected username %s, got username %s", creds.Username, username2)
+	}
+	if secret != "foobarbaz" {
```
fajpunk: D'oh! fixed.

osxkeychain/osxkeychain_darwin_test.go
```diff
@@ -54,6 +54,57 @@ func TestOSXKeychainHelper(t *testing.T) {
...
+		t.Fatalf("Error: expected secret %s, got secret %s", creds.Secret, secret2)
+	}
+
+	username3, secret3, err := helper.Get(loginServerURL)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if username != "foobar" {
```
fajpunk: D'oh! Fixed.

osxkeychain/osxkeychain_darwin_test.go
```diff
@@ -54,6 +54,57 @@ func TestOSXKeychainHelper(t *testing.T) {
...
+		t.Fatalf("Error: expected username %s, got username %s", creds.Username, username3)
+	}
+	if secret != "foobarbaz" {
```
fajpunk: D'oh! Fixed.

fajpunk:
Sorry for the late response :) I sometimes don't have access to a Mac, so I can't work on this all the time.

osxkeychain/osxkeychain_darwin.go
n4ss: `^\w+://` instead I'd say so that you match one or more `word chars` (a-z, A-Z, 0-9, '_') before a `://`

osxkeychain/osxkeychain_darwin.go
n4ss: Indeed!

osxkeychain/osxkeychain_darwin.go
fajpunk: Good call; updated.

vdemeester:
@fajpunk can you squash and sign your commits ? 👼

n4ss:
LGTM

fajpunk:
@vdemeester Signed and squashed.

n4ss:
@vdemeester are you ok with those changes so that I can merge?

n4ss:
edit: @thaJeztah just made some good comments, I'll request more changes sry :/

thaJeztah:
Yes, looking at this, I was wondering if we should craft our own RegEx; `url.Parse()` should already split out all the components, so we can check if `url.Scheme` is set/empty. There's quite a lot of permutations of URL's (think of weird cases such as `git+https://example.com`, or `https://user:pass@example.com`), for which a well-tested framework/library may be better.

In addition, we may want to consider if this is something to be handled by the *credentials_helper*, or by the Docker CLI (also see https://github.com/moby/moby/issues/33043#issuecomment-305483564).

Implementing it here results in the OS X keychain helper acting different than (e.g.) the Windows or Linux helper. Haven't thought this through entirely yet;

- if the keychain accepts values without scheme; should we modify the value, or accept as-is?
- if we _do_ want to consistently store these credentials, we should probably _pass_ them to the helper the way we want them to be stored

I agree that it's a bit surprising that the keychain _accepts_ the value, but mangles it afterwards (it should probably error-out if it's invalid)

fajpunk:
One problem with depending on url.Parse is that it doesn't parse protocol-less URLs correctly (or at least the way that we would expect them to be parsed here).  If given a URL without a protocol, url.Parse puts the domain in the path, which I think is the source of this bug:

```
package main

import (
	"fmt"
	"net/url"
)

func main() {
	wellFormed := "https://blah.com/something"
	poorlyFormed := "blah.com/something"
	
	goodUrl, _ := url.Parse(wellFormed)
	badUrl, _:= url.Parse(poorlyFormed)
	
	fmt.Println("Good scheme:", goodUrl.Scheme)
	fmt.Println("Bad scheme:", badUrl.Scheme)
	
	fmt.Println("")
	
	fmt.Println("Good path: ", goodUrl.Path)
	fmt.Println("Bad path:", badUrl.Path)
}

```Good scheme: https
Bad scheme: 

Good path:  /something
Bad path: blah.com/something

Program exited.```

https://play.golang.org/p/IAgk5TBuVa

**Comment by fajpunk** [comment]
I agree though that ideally, the credentials helper would only ever get passed well-formed URLs with protocols.

**Comment by thaJeztah** [comment]
Correct, in that case, scheme is not set, nor is hostname. We _could_ use _that_ information for validation possibly and/or rewriting (in Docker CLI) https://play.golang.org/p/uSIHiJ55ir

**Comment by fajpunk** [comment]
1. Do we want to close this PR in favor of implementing a fix in Docker CLI?
2. If we still want to implement a fix here, do we want to do the parse-then test-`IsAbs()`-then-modify-and-parse-again solution, or do we want to keep the regex-replace-and-then-parse solution?

I'm for the regex-replace-solution but modifying it back to a very liberal regex that just replaces everything up to `://`, given @thaJeztah's insightful comment about potentially non-alphanumeric protocols.  My reasoning is that `url.Parse` doesn't do what we expect (and I'd argue that it doesn't work correctly), so why would we depend on a portion of it's result while ditching another portion of it's result?

But I'm also totally bike-shedding here, and can see the value of depending on `IsAbs` too; I'm glad to implement whatever folks think is best.

What do you think @n4ss?  I'm happy to make whatever changes are necessary (assuming we want to keep the fix here).

**Comment by n4ss** [comment]
I think this should be fixed in both. Every level of abstraction should make sure that the URL is well formed imho, especially if some other tool decides to use creds-helper.

And apparently it looks like we only need to prepend `"//"` to the scheme-less URL in order to get a valid one.

**Comment by n4ss** [comment]
hey @fajpunk 

We discussed this parsing issue with @thaJeztah and he proposed the following snippet of code for `splitServer` which I like:

```go
func splitServer(serverURL string) (*C.struct_Server, error) {
    if !regexp.MustCompile(`^([a-zA-Z][-+.a-zA-Z0-9]+:)?//`).MatchString(serverURL) {
        serverURL = "//" + serverURL
    }
    u, err := url.Parse(serverURL)
    if err != nil {
        return nil, err
    }
    if u.Scheme != "" && u.Scheme != "https" && u.Scheme != "http" {
        return nil, errors.New("unsupported scheme: " + u.Scheme)
    }
    if url.Host() == "" {
        return nil, errors.New("no hostname in URL")
    }
    proto := C.kSecProtocolTypeHTTPS
    if u.Scheme == "http" {
        proto = C.kSecProtocolTypeHTTP
    }
    var port int
    if u.Port() != "" {
        port, err := strconv.Atoi(u.Port())
        if err != nil {
            return nil, err
        }
    }

    return &C.struct_Server{
        proto: C.SecProtocolType(proto),
        host:  C.CString(u.Host()),
        port:  C.uint(port),
        path:  C.CString(u.Path()),
    }, nil
}
```

This only replaces to `https` URLs with an empty scheme, or an `http`. I agree that we shouldn't replace other protocols. Would you mind modifying this function and the `ftp` test please as well as adding a test for something like `10.0.2.11:443` or something? I can do this worst case :) It seems like we'll need this for the next `docker` rc.

fajpunk:
Looks good to me!  I'm glad to do it, but won't have access to a Mac to
work on it until Thursday.  If that's OK, then I'll do it then.

n4ss:
@fajpunk I'll push on your branch on your repo if you're okay so that you remain the initial contributor to this issue and I don't want to steal credit :)
(we need it asap)

thaJeztah:
@n4ss oops, I had a branch I was working on. Needed some time to see if we could add useful tests https://github.com/docker/docker-credential-helpers/pull/72

fajpunk:
@n4ss, feel free to do whatever you need to do to get the fix in; push to my branch, close this and merge @thaJeztah's fix, whatever.  I don't care one bit about who gets credit for the fix (thought thanks for the thought), I'm just excited that it's going to be fixed :)

thaJeztah:
Thanks @fajpunk 👍

thaJeztah:
This was carried in https://github.com/docker/docker-credential-helpers/issues/72, thank you so much @fajpunk for starting the work on this!!

PR #62, opened 2017-05-22
Prevent invalid credentials: no missing server URL or username

n4ss:
We **should not** do operations on credential stores with invalid (empty) server URLs or usernames.

This causes credentials management issues in some platforms' helpers depending on the credentials management system implementation.

**This needs to be merged ASAP as next `docker` version is probably released tomorrow (05/23) morning CET.**

Signed-off-by: Nassim 'Nass' Eddequiouaq <eddequiouaq.nassim@gmail.com>


credentials/error.go
```diff
@@ -35,3 +42,73 @@ func IsErrCredentialsNotFound(err error) bool {
...
+func IsCredentialsMissingServerURL(err error) bool {
+	_, ok := err.(errCredentialsMissingServerURL)
+	return ok
+}
+
+// IsCredentialsMissingServerURLMessage returns true if the error
+// was caused by not having a credentials server URL when
+// required.
+func IsCredentialsMissingServerURLMessage(err string) bool {
+	return err == errCredentialsMissingServerURLMessage
```
thaJeztah: Is there a use for this utility method?

credentials/error.go
```diff
@@ -35,3 +42,73 @@ func IsErrCredentialsNotFound(err error) bool {
...
+	_, ok := err.(errCredentialsMissingUsername)
+	return ok
+}
+
+// IsCredentialsMissingUsernameMessage returns true if the error
+// was caused by not having a credentials username when
+// required.
+func IsCredentialsMissingUsernameMessage(err string) bool {
+	return err == errCredentialsMissingUsernameMessage
+}
```
thaJeztah: nit: missing newline at end of file

credentials/error.go
thaJeztah: (and the other `Is..Message` functions

n4ss: You're right, I forgot to add the error checks in the output logs from credentials store client.

Done in : https://github.com/docker/docker-credential-helpers/pull/62/commits/b69480e07432d492cbf5ef409dd80c7124b42796

credentials/error.go
n4ss: done!

aaronlehmann:
Are empty strings the only value that can trigger this issue? I know that the OS X keychain API returns the first "matching" credential, but I'm not sure whether "matching" means an exact match for all non-empty strings, or if prefix matching is done in some cases.

client/client.go
```diff
@@ -9,19 +9,39 @@ import (
 	"github.com/docker/docker-credential-helpers/credentials"
 )
 
+// isValidCredsMessage checks if 'msg' contains invalid credentials error message.
+// It returns whether the logs are free of invalid credentials errors and the error if it isn't.
+// error values can be errCredentialsMissingServerURL or errCredentialsMissingUsername.
+func isValidCredsMessage(msg string) (bool, error) {
```
aaronlehmann: Can we use a return code, or something similar, to convey this information? I really don't like the idea of baking an exact error string into the "protocol" here.

n4ss: That was being done before this PR.

You're totally right, I'd certainly prefer a return code too but this should go in another PR imho as I prefer not to alter anything that was here before and just add additional checks to fix the issue asap.

The helpers should use errors from `credentials` package so I guess that makes sense in the meantime we improve this.

n4ss:
@aaronlehmann Regarding your question, when trying to do get creds by the prefix of the stored credentials' `serverURL`, I got nothing. There is no documentation on the way the attributes are used to match existing keychain objects.

client/client.go
riyazdf: we probably could just have this function only return an `error` instead of `(bool, error)` - I don't see a use-case for them not matching

client/client.go
```diff
@@ -9,19 +9,39 @@ import (
...
 	buffer := new(bytes.Buffer)
-	if err := json.NewEncoder(buffer).Encode(credentials); err != nil {
+	if err := json.NewEncoder(buffer).Encode(creds); err != nil {
 		return err
 	}
 	cmd.Input(buffer)
 
 	out, err := cmd.Output()
 	if err != nil {
 		t := strings.TrimSpace(string(out))
```
riyazdf: for a followup PR: it might be useful to combine the  `cmd.Output()`, `strings.TrimSpace`, and `isValidCredsMessage` parsing into a shared function that all of these exported functions can use.
The same functionality seems to be replicated across `Get`, `Store`, `Erase`, etc.

credentials/credentials.go
```diff
@@ -100,13 +120,17 @@ func Get(helper Helper, reader io.Reader, writer io.Writer) error {
 	}
 
 	serverURL := strings.TrimSpace(buffer.String())
+	if len(serverURL) == 0 {
```
riyazdf: instead of this individual check, should we instead call `resp.isValid()` after forming it below?  That way we can also verify the username

client/client.go
n4ss: Indeed, added to my todo list :)

credentials/credentials.go
n4ss: the `Get` operation on a credentials store only needs a server URL, username is not needed, as the operation is the following:
- I have a repository `foobar.io`, get me the credentials for this repo so I can authenticate on it

client/client.go
```diff
@@ -9,19 +9,39 @@ import (
...
+
+		if err := isValidCredsMessage(t); err != nil {
+			return err
```
thaJeztah: It the error is one of these "special cases", only the bare message is returned, whereas other errors will show `"error storing credentials - err: ..."` Without that it may be less apparent in what stage the error occurred (storing or retrieving the credentials)

Is there a reason doing so? If we need to have access to the plain error message, perhaps we should consider using `errors.Wrap()`?

n4ss: I was actually following the "CredentialsNotFound" error check behavior, a follow-up PR fixing this is ready.

I prefer to have it separately though and not modify existing code.

n4ss: Are you ok with an additional parameter on `errCredentialsMissingUsername` and `errCredentialsMissingServerURL` errors that would include the operation in the error message?

credentials/credentials.go
riyazdf: got it.  But potentially, the username could create an invalid [`Credentials` struct in the code below](https://github.com/docker/docker-credential-helpers/pull/62/files#diff-9311ff8f32d06930be1eabeb53b517d6R134) right?  I would prefer to validate so that any invalid `Credentials` struct is rejected, in case the semantics of the `Get` operation change.

This can be a followup since I acknowledge the importance of this fix and that this is not an issue at present - I can help with the PR.

n4ss:
I'm squashing and merging, this will go in 17.06. Thanks all for reviewing.

PR #64, opened 2017-05-24
Caller of cred helpers would now see the stderr returns from cred helpers

shhsu:
As part of our design for Azure Docker Registry, we found that it would be useful (necessary for us) for cred manager to pipe it's stderr back to cli client. Hence this PR.

### DIFF
--- a/client/command.go
+++ b/client/command.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"io"
+	"os"
 	"os/exec"
 )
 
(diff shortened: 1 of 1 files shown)


shhsu:
@SajayAntony @DavidObando @sivagms

@friism, this is related to the redesign of https://github.com/docker/cli/pull/105

shhsu:
thanks @n4ss
Only those with write access can merge PR. Should I be expecting a merge when more people have approved the PR?

shhsu:
thanks @vdemeester

PR #67, opened 2017-05-30
Add version command

jeanlaurent:
This add a simple version switch to the CLI, to help determine which version is currently running.

Release is updated to read the version from sources.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: all deps osxkeychain secretservice test validate wincred
 
 TRAVIS_OS_NAME ?= linux
-VERSION = 0.5.1
+VERSION := $(shell grep 'const Version' credentials/version.go | awk -F'"' '{ print $$2 }')
 
 all: test
 
@@ -23,6 +23,7 @@ codesign: osxkeychain
 
 osxrelease: clean test codesign
 	mkdir -p release
+	@echo "\nPackaging version ${VERSION}\n"
 	cd bin && tar cvfz ../release/docker-credential-osxkeychain-v$(VERSION)-amd64.tar.gz docker-credential-osxkeychain
 
 secretservice:
--- a/Makefile
+++ b/Makefile
@@ -16,12 +16,12 @@ osxkeychain:
 	mkdir bin
 	go build -ldflags -s -o bin/docker-credential-osxkeychain osxkeychain/cmd/main_darwin.go
 
-codesign: osxkeychain
+osxcodesign: osxkeychain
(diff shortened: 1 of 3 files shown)


jeanlaurent:
@thaJeztah done in e69fae4
git sha1 would be cool, but require more work, maybe worth an issue ?

Makefile
```diff
@@ -1,7 +1,7 @@
 .PHONY: all deps osxkeychain secretservice test validate wincred
 
 TRAVIS_OS_NAME ?= linux
-VERSION = 0.5.1
+VERSION := $(shell grep const credentials/version.go | awk -F'"' '{ print $$2 }')
```
n4ss: `grep "const Version"` and `print Version $$2`

n4ss: just in case we have more const variables in this file.

Makefile
```diff
@@ -16,13 +16,14 @@ osxkeychain:
...
 	go build -ldflags -s -o bin/docker-credential-osxkeychain osxkeychain/cmd/main_darwin.go
 
-codesign: osxkeychain
+osxcodesign: osxkeychain
 	$(eval SIGNINGHASH = $(shell security find-identity -v -p codesigning | grep "Developer ID Application: Docker Inc" | cut -d ' ' -f 4))
 	xcrun -log codesign -s $(SIGNINGHASH) --force --verbose bin/docker-credential-osxkeychain
 	xcrun codesign --verify --deep --strict --verbose=2 --display bin/docker-credential-osxkeychain
 
-osxrelease: clean test codesign
+osxrelease: clean vet_osx lint fmt test osxcodesign
```
n4ss: 👍 good catch, sorry for not checking on my last PR and not making change before..

n4ss:
I prefer the version tag as it is right now personally.

LGTM after the nit "const Version" comment.

credentials/version.go
```diff
@@ -1,4 +1,4 @@
 package credentials
 
-// Version holds a string describing the current version
-const Version = "0.5.1"
+// Version holds a string describing the current versio
```
thaJeztah: typo: s/versio/version/

Makefile
jeanlaurent: done

credentials/version.go
jeanlaurent: fixed

Makefile
```diff
@@ -1,7 +1,7 @@
 .PHONY: all deps osxkeychain secretservice test validate wincred
 
 TRAVIS_OS_NAME ?= linux
-VERSION = 0.5.1
+VERSION := $(shell grep 'const Version' credentials/version.go | awk -F'"' '{ print $$2 }')
```
thaJeztah: For a follow-up, perhaps we should consider having a `VERSION` file in the repo

PR #69, opened 2017-05-31
A new entry point for calling the cred helpers that allow passing environment variables

shhsu:
This change introduce a new entry point for credential helpers. The caller can now pass environment variable so that the credential helper can be aware of them.

The motivation of this feature is that Azure Container Registry is developing a credential helper that wants to know which docker file is currently active. Note that when `docker` command was called with `--config` flag, there is currently no way for docker to pass that information to the credential helper.

This feature will allow changes to be made in docker cli so it could, for example set the `DOCKER_CONFIG` environmental variable before calling the credential helper.

### DIFF
--- a/client/command.go
+++ b/client/command.go
@@ -1,6 +1,7 @@
 package client
 
 import (
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


shhsu:
@sajayantony @davidobando

n4ss:
I like this, too bad we weren't doing this before :)

shhsu:
Updated as requested

DavidObando:
Ping.

n4ss:
@vdemeester @thaJeztah WDYT?

n4ss:
LGTM

PR #72, opened 2017-06-14
Fix storing URLs without scheme

thaJeztah:
If secrets are stored without specifying a scheme
(https://), the keychain-helper would interpret the
hostname as _path_, causing lookup of secrets to fail.

This patch makes sure that a scheme is added (if missing).

If no scheme is specified, https:// is used as a default.

### DIFF
--- a/osxkeychain/osxkeychain_darwin.go
+++ b/osxkeychain/osxkeychain_darwin.go
@@ -15,6 +15,7 @@ import (
 	"unsafe"
 
 	"github.com/docker/docker-credential-helpers/credentials"
+	"strings"
 )
 
 // errCredentialsNotFound is the specific error message returned by OS X
@@ -173,18 +174,16 @@ func freeServer(s *C.struct_Server) {
 // before parsing. This prevents the hostname being used as path,
 // and the credentials being stored without host.
 func parseURL(serverURL string) (*url.URL, error) {
+	// Check if serverURL has a valid scheme, otherwise add `//` as scheme.
+	if !strings.Contains(serverURL, "://") && !strings.HasPrefix(serverURL, "//") {
+		serverURL = "//" + serverURL
+	}
+
 	u, err := url.Parse(serverURL)
 	if err != nil {
 		return nil, err
 	}
-	// Check if serverURL has a valid scheme, otherwise add `//` as scheme.
-	// Without a scheme, the hostname is seen as "path" by `url.Parse()`
-	if u.Scheme == "" {
-		u, err = url.Parse("//" + serverURL)
-		if err != nil {
-			return nil, err
-		}
-	}
+
 	if u.Scheme != "" && u.Scheme != "https" && u.Scheme != "http" {
 		return nil, errors.New("unsupported scheme: " + u.Scheme)
 	}
--- a/osxkeychain/osxkeychain_darwin.go
+++ b/osxkeychain/osxkeychain_darwin.go
@@ -12,10 +12,10 @@ import (
 	"errors"
 	"net/url"
 	"strconv"
+	"strings"
 	"unsafe"
 
 	"github.com/docker/docker-credential-helpers/credentials"
-	"strings"
 )
 
 // errCredentialsNotFound is the specific error message returned by OS X
@@ -174,7 +174,7 @@ func freeServer(s *C.struct_Server) {
 // before parsing. This prevents the hostname being used as path,
 // and the credentials being stored without host.
 func parseURL(serverURL string) (*url.URL, error) {
-	// Check if serverURL has a valid scheme, otherwise add `//` as scheme.
+	// Check if serverURL has a scheme, otherwise add `//` as scheme.
 	if !strings.Contains(serverURL, "://") && !strings.HasPrefix(serverURL, "//") {
 		serverURL = "//" + serverURL
 	}
--- a/osxkeychain/osxkeychain_darwin.go
+++ b/osxkeychain/osxkeychain_darwin.go
@@ -12,7 +12,6 @@ import (
 	"errors"
 	"net/url"
 	"strconv"
-	"strings"
 	"unsafe"
 
 	"github.com/docker/docker-credential-helpers/credentials"
@@ -135,30 +134,26 @@ func (h Osxkeychain) List() (map[string]string, error) {
 }
 
 func splitServer(serverURL string) (*C.struct_Server, error) {
-	u, err := url.Parse(serverURL)
+	u, err := parseURL(serverURL)
 	if err != nil {
 		return nil, err
 	}
 
-	hostAndPort := strings.Split(u.Host, ":")
-	host := hostAndPort[0]
+	proto := C.kSecProtocolTypeHTTPS
+	if u.Scheme == "http" {
+		proto = C.kSecProtocolTypeHTTP
+	}
 	var port int
-	if len(hostAndPort) == 2 {
-		p, err := strconv.Atoi(hostAndPort[1])
+	if u.Port() != "" {
+		port, err = strconv.Atoi(u.Port())
 		if err != nil {
 			return nil, err
 		}
-		port = p
-	}
-
-	proto := C.kSecProtocolTypeHTTPS
-	if u.Scheme != "https" {
-		proto = C.kSecProtocolTypeHTTP
 	}
 
 	return &C.struct_Server{
 		proto: C.SecProtocolType(proto),
-		host:  C.CString(host),
+		host:  C.CString(u.Hostname()),
 		port:  C.uint(port),
 		path:  C.CString(u.Path),
 	}, nil
@@ -167,4 +162,35 @@ func splitServer(serverURL string) (*C.struct_Server, error) {
 func freeServer(s *C.struct_Server) {
 	C.free(unsafe.Pointer(s.host))
 	C.free(unsafe.Pointer(s.path))
+}
+
+// parseURL parses and validates a given serverURL to an url.URL, and
+// returns an error if validation failed. Querystring parameters are
+// omitted in the resulting URL, because they are not used in the helper.
+//
+// If serverURL does not have a valid scheme, `//` is used as scheme
+// before parsing. This prevents the hostname being used as path,
+// and the credentials being stored without host.
+func parseURL(serverURL string) (*url.URL, error) {
+	u, err := url.Parse(serverURL)
+	if err != nil {
+		return nil, err
+	}
+	// Check if serverURL has a valid scheme, otherwise add `//` as scheme.
+	// Without a scheme, the hostname is seen as "path" by `url.Parse()`
+	if u.Scheme == "" {
+		u, err = url.Parse("//" + serverURL)
+		if err != nil {
+			return nil, err
+		}
+	}
+	if u.Scheme != "" && u.Scheme != "https" && u.Scheme != "http" {
+		return nil, errors.New("unsupported scheme: " + u.Scheme)
+	}
+	if u.Hostname() == "" {
+		return nil, errors.New("no hostname in URL")
+	}
+
+	u.RawQuery = ""
+	return u, nil
 }
--- a/osxkeychain/osxkeychain_darwin.go
+++ b/osxkeychain/osxkeychain_darwin.go
@@ -144,8 +144,9 @@ func splitServer(serverURL string) (*C.struct_Server, error) {
 		proto = C.kSecProtocolTypeHTTP
 	}
 	var port int
-	if u.Port() != "" {
-		port, err = strconv.Atoi(u.Port())
+	p := getPort(u)
+	if p != "" {
+		port, err = strconv.Atoi(p)
 		if err != nil {
 			return nil, err
 		}
@@ -153,7 +154,7 @@ func splitServer(serverURL string) (*C.struct_Server, error) {
 
 	return &C.struct_Server{
 		proto: C.SecProtocolType(proto),
-		host:  C.CString(u.Hostname()),
+		host:  C.CString(getHostname(u)),
 		port:  C.uint(port),
 		path:  C.CString(u.Path),
 	}, nil
@@ -187,7 +188,7 @@ func parseURL(serverURL string) (*url.URL, error) {
 	if u.Scheme != "" && u.Scheme != "https" && u.Scheme != "http" {
 		return nil, errors.New("unsupported scheme: " + u.Scheme)
 	}
-	if u.Hostname() == "" {
+	if getHostname(u) == "" {
 		return nil, errors.New("no hostname in URL")
 	}
 
--- a/osxkeychain/osxkeychain_darwin_test.go
+++ b/osxkeychain/osxkeychain_darwin_test.go
@@ -1,6 +1,8 @@
 package osxkeychain
 
 import (
+	"errors"
+	"fmt"
 	"github.com/docker/docker-credential-helpers/credentials"
 	"testing"
 )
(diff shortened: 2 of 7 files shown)


osxkeychain/osxkeychain_darwin_test.go
```diff
@@ -54,6 +56,157 @@ func TestOSXKeychainHelper(t *testing.T) {
...
+		readURL  string
+	}{
+		// stored as https, retrieved using http
+		{"https://foobar.docker.io:2376", "http://foobar.docker.io:2376"},
+
+		// stored as http, retrieved using https
+		{"http://foobar.docker.io:2376", "https://foobar.docker.io:2376"},
+
+		// same: stored as http, retrieved without a scheme specified (hence, using the default https://)
+		// TODO is this desired behavior?
```
thaJeztah: Wasn't sure about this one

osxkeychain/osxkeychain_darwin_test.go
```diff
@@ -54,6 +56,157 @@ func TestOSXKeychainHelper(t *testing.T) {
...
+		{"http://foobar.docker.io", "foobar.docker.io:5678"},
+
+		// non-matching ports
+		{"https://foobar.docker.io:1234", "https://foobar.docker.io:5678"},
+
+		// non-matching ports
+		// TODO is this desired behavior? The other way round does work
```
thaJeztah: Same here

osxkeychain/osxkeychain_darwin.go
```diff
@@ -168,3 +164,25 @@ func freeServer(s *C.struct_Server) {
 	C.free(unsafe.Pointer(s.host))
 	C.free(unsafe.Pointer(s.path))
 }
+
+// parseURL parses and validates a given serverURL to an url.URL, and
+// returns an error if validation failed. Querystring parameters are
+// omitted in the resulting URL, because they are not used in the helper
+func parseURL(serverURL string) (*url.URL, error) {
```
thaJeztah: Splitted this to a separate function, so that we can test independently 😄

osxkeychain/osxkeychain_darwin.go
```diff
@@ -168,3 +164,25 @@ func freeServer(s *C.struct_Server) {
...
+		return nil, err
+	}
+	if u.Scheme != "" && u.Scheme != "https" && u.Scheme != "http" {
+		return nil, errors.New("unsupported scheme: " + u.Scheme)
+	}
+	if u.Hostname() == "" {
+		return nil, errors.New("no hostname in URL")
+	}
+
+	u.RawQuery = ""
```
thaJeztah: Explicitly strip query-string, because we're not storing it, but let me know if you don't want this 👍

osxkeychain/osxkeychain_darwin_test.go
```diff
@@ -54,6 +56,161 @@ func TestOSXKeychainHelper(t *testing.T) {
...
+		readURL  string
+	}{
+		// stored with port, retrieved without
+		{"https://foobar.docker.io:2376", "https://foobar.docker.io"},
+
+		// stored as https, retrieved without scheme
+		{"https://foobar.docker.io:2376", "foobar.docker.io"},
+
+		// stored with path, retrieved without
+		// TODO is this the desired behavior?
```
thaJeztah: Wasn't sure about this either

osxkeychain/osxkeychain_darwin.go
```diff
@@ -168,3 +164,25 @@ func freeServer(s *C.struct_Server) {
 	C.free(unsafe.Pointer(s.host))
 	C.free(unsafe.Pointer(s.path))
 }
+
+// parseURL parses and validates a given serverURL to an url.URL, and
+// returns an error if validation failed. Querystring parameters are
+// omitted in the resulting URL, because they are not used in the helper
+func parseURL(serverURL string) (*url.URL, error) {
+	if !regexp.MustCompile(`^([a-zA-Z][-+.a-zA-Z0-9]+:)?//`).MatchString(serverURL) {
```
n4ss: nit: can we add a comment to know what this regex checks?

osxkeychain/osxkeychain_darwin.go
n4ss: Very good catch!

osxkeychain/osxkeychain_darwin.go
```diff
@@ -135,30 +135,26 @@ func (h Osxkeychain) List() (map[string]string, error) {
...
-
-	proto := C.kSecProtocolTypeHTTPS
-	if u.Scheme != "https" {
-		proto = C.kSecProtocolTypeHTTP
 	}
 
 	return &C.struct_Server{
 		proto: C.SecProtocolType(proto),
-		host:  C.CString(host),
+		host:  C.CString(u.Hostname()),
```
n4ss: Is `u.Hostname()` different from `u.Host` ?

osxkeychain/osxkeychain_darwin_test.go
n4ss: Path is not kept in the keychain, this is the desired behavior AAIK 👍

osxkeychain/osxkeychain_darwin_test.go
```diff
@@ -54,6 +56,161 @@ func TestOSXKeychainHelper(t *testing.T) {
...
+
+	for _, te := range tests {
+		u, err := parseURL(te.url)
+
+		if te.err == nil && err != nil {
+			t.Errorf("Error: failed to parse URL %q: %s", te.url, err)
+			continue
+		}
+		if te.err != nil && err == nil {
+			t.Errorf("Error: expected parsing to fail for URL %q", te.url)
```
n4ss: Can we include the expected error message of `te.err`?

osxkeychain/osxkeychain_darwin_test.go
```diff
@@ -54,6 +56,161 @@ func TestOSXKeychainHelper(t *testing.T) {
...
+		}
+		if _, _, err := helper.Get(te.readURL); err != nil {
+			t.Errorf("Error: failed to read secret for URL %q using %q", te.storeURL, te.readURL)
+		}
+		helper.Delete(te.storeURL)
+	}
+}
+
+// TestOSXKeychainHelperRetrieveStrict verifies that only matching secrets are
+// returned. Secrets
```
n4ss: s/"Secret"//

osxkeychain/osxkeychain_darwin_test.go
n4ss: This one looks good to me

osxkeychain/osxkeychain_darwin_test.go
n4ss: Not sure here either, I'd drop this test.

osxkeychain/osxkeychain_darwin_test.go
```diff
@@ -54,6 +56,161 @@ func TestOSXKeychainHelper(t *testing.T) {
...
+
+	helper := Osxkeychain{}
+	for i, te := range tests {
+		c := &credentials.Credentials{
+			ServerURL: te.url,
+			Username:  fmt.Sprintf("user-%d", i),
+			Secret:    fmt.Sprintf("secret-%d", i),
+		}
+
+		if err := helper.Add(c); err != nil {
```
n4ss: `helper.Delete` before testing and `defer helper.Delete` in order to remove all the added credentials, otherwise it generates false positives.

osxkeychain/osxkeychain_darwin_test.go
```diff
@@ -54,6 +56,161 @@ func TestOSXKeychainHelper(t *testing.T) {
...
+		{"https://foobar.docker.io", "https://foobar.docker.io:5678"},
+
+		// non-matching paths
+		{"https://foobar.docker.io:1234/one/two", "https://foobar.docker.io:1234/five/six"},
+	}
+
+	helper := Osxkeychain{}
+	for _, te := range tests {
+		c := &credentials.Credentials{ServerURL: te.storeURL, Username: "hello", Secret: "world"}
+		if err := helper.Add(c); err != nil {
```
n4ss: Same thing here for the Delete -> Add -> defer Delete

osxkeychain/osxkeychain_darwin_test.go
```diff
@@ -54,6 +56,161 @@ func TestOSXKeychainHelper(t *testing.T) {
...
+
+		// stored with path, retrieved without
+		// TODO is this the desired behavior?
+		{"https://foobar.docker.io:1234/one/two", "https://foobar.docker.io:1234"},
+	}
+
+	helper := Osxkeychain{}
+	for _, te := range tests {
+		c := &credentials.Credentials{ServerURL: te.storeURL, Username: "hello", Secret: "world"}
+		if err := helper.Add(c); err != nil {
```
n4ss: Same thing here for the Delete -> Add -> defer Delete

osxkeychain/osxkeychain_darwin_test.go
n4ss: I did comment backward for this one sorry, added references to other spots where it should be done too.

osxkeychain/osxkeychain_darwin.go
thaJeztah: `u.Hostname()` strips the port; https://golang.org/src/net/url/url.go?s=28154:28185#L997

osxkeychain/osxkeychain_darwin.go
thaJeztah: Yup, makes sense

n4ss:
LGTM! travisCI is failing because it compiles with Golang 1.6, I don't have the rights to update I think.

@vdemeester can you review this please? ^_^

n4ss:
Follow up issue on this on the CLI side: https://github.com/docker/cli/issues/190

We now need to fix the behavior of stripping the protocol when a user adds it to the URL.

Since we'll default to HTTPS, they'll have explicitly specify protocol: `docker login http://...` to use an insecure connection.

osxkeychain/osxkeychain_darwin_test.go
thaJeztah: For future reference; this was addressed through this part above;

```go
// Clean store before testing.
for _, te := range tests {
	helper.Delete(te.url)
}
```

thaJeztah:
Hm, looks like it fails on Go 1.7 as well; were these options added in 1.8?

```
ok  	github.com/docker/docker-credential-helpers/credentials	0.224s
# github.com/docker/docker-credential-helpers/osxkeychain
osxkeychain/osxkeychain_darwin.go:148: u.Port undefined (type *url.URL has no field or method Port)
osxkeychain/osxkeychain_darwin.go:149: u.Port undefined (type *url.URL has no field or method Port)
osxkeychain/osxkeychain_darwin.go:157: u.Hostname undefined (type *url.URL has no field or method Hostname)
osxkeychain/osxkeychain_darwin.go:188: u.Hostname undefined (type *url.URL has no field or method Hostname)
# github.com/docker/docker-credential-helpers/osxkeychain
osxkeychain/osxkeychain_darwin.go:148: u.Port undefined (type *url.URL has no field or method Port)
osxkeychain/osxkeychain_darwin.go:149: u.Port undefined (type *url.URL has no field or method Port)
osxkeychain/osxkeychain_darwin.go:157: u.Hostname undefined (type *url.URL has no field or method Hostname)
osxkeychain/osxkeychain_darwin.go:188: u.Hostname undefined (type *url.URL has no field or method Hostname)
FAIL	github.com/docker/docker-credential-helpers/osxkeychain [build failed]
make: *** [test] Error 2
```

thaJeztah:
Ugh looks like it was added in Go 1.8 https://golang.org/doc/go1.8

screen shot 2017-06-14 at 18 47 57

osxkeychain/osxkeychain_darwin.go
```diff
@@ -168,3 +164,31 @@ func freeServer(s *C.struct_Server) {
...
+	}
+	if u.Scheme != "" && u.Scheme != "https" && u.Scheme != "http" {
+		return nil, errors.New("unsupported scheme: " + u.Scheme)
+	}
+	if u.Hostname() == "" {
+		return nil, errors.New("no hostname in URL")
+	}
+
+	u.RawQuery = ""
+	return u, nil
```
dnephin: Instead of striping the query, and checking the scheme above what do you think about returning a different struct here instead of `url.Url`. It seems we're not really using much of the url.

```go
type url struct {
    proto string
    host string
    port string
    path string
}
```

osxkeychain/osxkeychain_darwin.go
```diff
@@ -168,3 +164,31 @@ func freeServer(s *C.struct_Server) {
...
+//
+// If serverURL does not have a valid scheme, `//` is used as scheme
+// before parsing. This prevents the hostname being used as path,
+// and the credentials being stored without host.
+func parseURL(serverURL string) (*url.URL, error) {
+	// Check if serverURL has a valid scheme, otherwise add `//` as scheme.
+	// Without a scheme, the hostname is seen as "path" by `url.Parse()`
+	if !regexp.MustCompile(`^([a-zA-Z][-+.a-zA-Z0-9]+:)?//`).MatchString(serverURL) {
+		serverURL = "//" + serverURL
+	}
```
dnephin: I think this could call `url.Parse()` and if `u.Scheme == ""`, then prepend `//` and call it again on the new string.

osxkeychain/osxkeychain_darwin.go
```diff
@@ -168,3 +164,31 @@ func freeServer(s *C.struct_Server) {
...
+// returns an error if validation failed. Querystring parameters are
+// omitted in the resulting URL, because they are not used in the helper.
+//
+// If serverURL does not have a valid scheme, `//` is used as scheme
+// before parsing. This prevents the hostname being used as path,
+// and the credentials being stored without host.
+func parseURL(serverURL string) (*url.URL, error) {
+	// Check if serverURL has a valid scheme, otherwise add `//` as scheme.
+	// Without a scheme, the hostname is seen as "path" by `url.Parse()`
+	if !regexp.MustCompile(`^([a-zA-Z][-+.a-zA-Z0-9]+:)?//`).MatchString(serverURL) {
```
tiborvass: no biggie, but if you update the PR, you could put this mustcompile in a package level var

osxkeychain/osxkeychain_darwin.go
```diff
@@ -135,30 +135,26 @@ func (h Osxkeychain) List() (map[string]string, error) {
...
-	hostAndPort := strings.Split(u.Host, ":")
-	host := hostAndPort[0]
+	proto := C.kSecProtocolTypeHTTPS
+	if u.Scheme == "http" {
+		proto = C.kSecProtocolTypeHTTP
+	}
 	var port int
-	if len(hostAndPort) == 2 {
-		p, err := strconv.Atoi(hostAndPort[1])
+	if u.Port() != "" {
```
tiborvass: These go1.8 Port() and Hostname() methods are only useful for supporting ipv6, based on what I've read from their code. We could argue it's out of scope for this PR. Or we should provide pre-1.8 functions with `!go1.8` buildtags

osxkeychain/osxkeychain_darwin.go
tiborvass: Good idea!

osxkeychain/osxkeychain_darwin.go
tiborvass: We're using the Port() and Hostname() methods

dnephin: Right, but those values could be written to this new struct before running. Either way works fine.

tiborvass:
ping @dnephin @thaJeztah

osxkeychain/osxkeychain_darwin.go
thaJeztah: Oh, funny, that was my original proposal; https://github.com/docker/docker-credential-helpers/pull/55#issuecomment-305507058 / https://play.golang.org/p/uSIHiJ55ir

osxkeychain/osxkeychain_darwin.go
thaJeztah: I was thinking of that first, but decided to keep it simple

osxkeychain/osxkeychain_darwin.go
```diff
@@ -168,3 +164,34 @@ func freeServer(s *C.struct_Server) {
...
+// before parsing. This prevents the hostname being used as path,
+// and the credentials being stored without host.
+func parseURL(serverURL string) (*url.URL, error) {
+	u, err := url.Parse(serverURL)
+	if err != nil {
+		return nil, err
+	}
+	// Check if serverURL has a valid scheme, otherwise add `//` as scheme.
+	// Without a scheme, the hostname is seen as "path" by `url.Parse()`
+	if u.Scheme == "" {
```
thaJeztah: This may not actually work; `localhost:443` will be parsed as `scheme: localhost, host: 443`; https://play.golang.org/p/QRE7pqxjx1

thaJeztah: To simplify, we can use

```go
if !strings.Contains(serverURL, "://") && ! strings.HasPrefix(serverURL, "//") {
    serverURL = "//" + serverURL
}
```

tiborvass:
Another note: this should be a fix for all OSes and not specific to OSX. Fixing.

osxkeychain/osxkeychain_darwin.go
n4ss: Sounds good

n4ss:
@tiborvass at the wrapper level? Then, we'll have to warn every folk implementing a credential helper.

tiborvass:
LGTM

osxkeychain/osxkeychain_darwin.go
```diff
@@ -12,10 +12,10 @@ import (
 	"errors"
 	"net/url"
 	"strconv"
-	"strings"
 	"unsafe"
 
 	"github.com/docker/docker-credential-helpers/credentials"
+	"strings"
```
thaJeztah: Is this correct?

osxkeychain/osxkeychain_darwin.go
```diff
@@ -168,3 +165,32 @@ func freeServer(s *C.struct_Server) {
...
+
+// parseURL parses and validates a given serverURL to an url.URL, and
+// returns an error if validation failed. Querystring parameters are
+// omitted in the resulting URL, because they are not used in the helper.
+//
+// If serverURL does not have a valid scheme, `//` is used as scheme
+// before parsing. This prevents the hostname being used as path,
+// and the credentials being stored without host.
+func parseURL(serverURL string) (*url.URL, error) {
+	// Check if serverURL has a valid scheme, otherwise add `//` as scheme.
```
thaJeztah: can you remove "valid" here in the comment? This is just a detection is a "possible" scheme is present

osxkeychain/osxkeychain_darwin.go
n4ss: According to `gofmt` it is :/

osxkeychain/osxkeychain_darwin_test.go
```diff
@@ -1,6 +1,8 @@
 package osxkeychain
 
 import (
+	"errors"
+	"fmt"
 	"github.com/docker/docker-credential-helpers/credentials"
 	"testing"
```
thaJeztah: hm, not sure what formatting is used in this repo, but ideally would be;

```go
import (
	"errors"
	"fmt"
	"testing"

	"github.com/docker/docker-credential-helpers/credentials"
)
```

Just a nit, so not a blocker

n4ss: I'll investigate that for a next PR :)

osxkeychain/osxkeychain_darwin.go
thaJeztah: Ah, I guess it's because docker/docker follows the `goimport` conventions, in which case they are [grouped, and sorted](https://groups.google.com/forum/#!topic/golang-nuts/Ir0-hKNkEkQ), but `gofmt` won't complain if they were not grouped.

> gofmt follows a simple rule, which is to sort each group of imports, 
> where different groups are separated by a blank line. 
>
> goimports also follows a simple rule, which is to put stdlib imports 
> in one group and non-stdlib imports in a different group.

dnephin:
Someone just linked me this: https://github.com/goware/urlx

I think this is exactly what we want for the url parsing code.

thaJeztah:
@dnephin that looks neat indeed. Only issue I see is that it defaults to http, but perhaps we can make a proposal to make it configurable, or have a function that allows passing in defaults for missing components

PR #79, opened 2017-06-21
README update to add homebrew installation help

danshumaker:
This tripped me up for a few minutes, so I thought it would be helpful to add to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@ docker-credential-helpers is a suite of programs to use native stores to keep Do
 
 Go to the [Releases](https://github.com/docker/docker-credential-helpers/releases) page and download the binary that works better for you. Put that binary in your `$PATH`, so Docker can find it.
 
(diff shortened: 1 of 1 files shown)


n4ss:
LGTM

/cc @vdemeester

promiseofcake:
This has actually been changed to `docker-credential-helper`. Via: https://github.com/Homebrew/homebrew-core/blob/master/Formula/docker-credential-helper.rb

guillaumerose:
@danshumaker Can you update the PR with the correct formula name so that I can merge ?

Thanks and sorry for the delay

danshumaker:
@guilamarose not sure i know what you mean. Seems like from promiseofcake that my change is not necessary. I said the formula was docker-credential-osxhelper but it looks like docker-credential-helper is now the right one. Haven’t tested it in a couple years.

Tyson1986:
For me 
`brew install docker-credential-helper`
works, and 
`brew install docker-credential-osxhelper`
does not work.

PR #81, opened 2017-08-10
Add `pass` backend and debian packaging for both -secretservice and -pass backends

tych0:
On the road to securing passwords by default in some way, we need two things: packaging for whatever password backend we use, and a password backend that doesn't require X by default.

This commit attempts to provide both of those things, using the `pass` utility, which uses gpg directly, instead of dbus and an external daemon. It does, however, require users to have configured `pass` with their key correctly.

I'm not sure what to do about really trying to get packaging in the hands of users. For docker there are a lot of various combinations that we generate packages for; I suppose we need to replicate that here?

cc @n4ss


pass/pass_linux_test.go
```diff
@@ -0,0 +1,80 @@
...
+	helper.Add(creds)
+
+	credsList, err := helper.List()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for server, username := range credsList {
+		if !strings.Contains(server, "2376") &&
+			!strings.Contains(creds.ServerURL, "9999") {
```
n4ss: did you mean:
```
if !strings.Contains(server, "2376") || !strings.Contains(server, "9999") { ...
```

pass/pass_linux_test.go
```diff
@@ -0,0 +1,80 @@
...
+		if u != username {
+			t.Fatalf("invalid username %s", u)
+		}
+
+		if s != "isthebestmeshuggahalbum" {
+			t.Fatalf("invalid secret: %s", s)
+		}
+	}
+
+	err = helper.Delete("https://foobar.docker.io:9999/v2")
```
n4ss: Can we add these `Delete` inside the above for loop? that way we can add more test cases and not have duplicate code outside of it.

pass/pass_linux.go
```diff
@@ -0,0 +1,209 @@
...
+		return errors.New("pass store is unitialized")
+	}
+
+	if creds == nil {
+		return errors.New("missing credentials")
+	}
+
+	encoded := base64.URLEncoding.EncodeToString([]byte(creds.ServerURL))
+
+	_, err := runPass(creds.Secret, "insert", "-f", "-m", path.Join(PASS_FOLDER, encoded, creds.Username))
```
tych0: I don't think so; URLEncoding explicitly does not encode to slashes, which is why I chose it: https://golang.org/src/encoding/base64/base64.go#L35

pass/pass_linux.go
```diff
@@ -0,0 +1,209 @@
...
+
+	servers, err := listPassDir()
+	if err != nil {
+		return nil, err
+	}
+
+	resp := map[string]string{}
+
+	for _, server := range servers {
+		if !server.IsDir() {
```
tych0: No, I use it twice, once to list directories, and the other to figure out usernames, which are indicated by the file name.

pass/pass_linux_test.go
tych0: Yes, both the || and the fix for `s/creds.ServerURL/server`, thanks!

pass/pass_linux_test.go
tych0: It changes the semantics of the test slightly, but I don't mind.

pass/pass_linux_test.go
tych0: Oh, sorry, I don't mean the `||` bit. This is a testing of failure conditions, so it should be &&. The string should be one of 2376 or 9999, which means the failure condition is `!(2376 or 9999) => !2376 && !9999`; I'll switch it around to `!(2376 or 9999)`, though.

tych0:
Thanks @n4ss, I think I've addressed it all.

tych0:
Also, go back to vacation :)

pass/pass_linux.go
n4ss: Oh my bad didn't see that, very good then!

pass/pass_linux.go
n4ss: Makes sense, I read it as "listPassDirs" instead of "listPassDir"...

n4ss:
LGTM

runcom:
LGTM

PR #83, opened 2017-08-14
pass backend: expose pass initialized flag

tych0:
This way clients don't have to run their own checks, they can just use the
library's detection.


n4ss:
LGTM, good catch!

PR #84, opened 2017-08-17
return "" instead of an error when pass isn't present

tych0:
It turns out the cred helpers protocol is to return "" and not an error
when a credential isn't present, so let's do that.


tych0:
Ping, can we merge this?

dnephin:
Not really sure what the policy is on this repo, but generally 2 LGTM is enough. I think you can merge , yes?

tych0:
No, I don't have +w here. Thanks though!

PR #89, opened 2017-08-28
pass: better initialization check

tych0:
See comment for details, but basically, let's actually use pass to check
that it works.


n4ss:
LGTM!

/cc @vdemeester

Issue #96, opened 2017-11-02
docker runs "pass" always?!


blueyed:
I've noticed that just running `docker` will execute `pass`:

```
% strace -f -e file docker
execve("/usr/bin/docker", ["docker"], 0x7ffe510b2898 /* 82 vars */) = 0
access("/etc/ld.so.preload", R_OK)      = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/etc/ld.so.cache", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/usr/lib/libpthread.so.0", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/usr/lib/libltdl.so.7", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/usr/lib/libc.so.6", O_RDONLY|O_CLOEXEC) = 3
openat(AT_FDCWD, "/usr/lib/libdl.so.2", O_RDONLY|O_CLOEXEC) = 3
strace: Process 8605 attached
strace: Process 8606 attached
strace: Process 8607 attached
strace: Process 8608 attached
[pid  8604] readlinkat(AT_FDCWD, "/proc/self/exe", "/usr/bin/docker", 128) = 15
[pid  8604] openat(AT_FDCWD, "/proc/sys/net/core/somaxconn", O_RDONLY|O_CLOEXEC) = 3
[pid  8604] stat("/home/user/bin/foo/bin/pass", 0xc4200a0fa8) = -1 ENOENT (No such file or directory)
…
[pid  8604] stat("/usr/bin/pass", {st_mode=S_IFREG|0755, st_size=24974, ...}) = 0
[pid  8604] openat(AT_FDCWD, "/dev/null", O_RDONLY|O_CLOEXEC) = 3
[pid  8604] openat(AT_FDCWD, "/dev/null", O_WRONLY|O_CLOEXEC) = 5
[pid  8604] openat(AT_FDCWD, "/dev/null", O_WRONLY|O_CLOEXEC) = 6
strace: Process 8609 attached
[pid  8609] execve("/usr/bin/pass", ["pass"], 0xc4200b8840 /* 82 vars */) = 0
```

I've tried to configure `credsStore`, but `~/.docker/config.json` gets only read later.

I think this should be avoided?!

Docker version 17.10.0-ce, build f4ffd2511c
Arch Linux

deeky666:
[...]

PR #106, opened 2018-06-18
Remove the automatic call for `pass` in `docker-credential-pass`.

vdemeester:
By having those calls in `init`, they are executed for any command,
and for any external code that depends on this package.

Replacing it by a `IsPresent` function and an internal `isInitialized`
function that will do the calls.

Fix https://github.com/docker/docker-credential-helpers/issues/96

cc @n4ss @thaJeztah

Signed-off-by: Vincent Demeester <vincent@sbr.pm>


pass/pass_linux.go
```diff
@@ -20,29 +20,13 @@ import (
...
-
-	stored, err := runPass("", "show", name)
-	PassInitialized = err == nil && stored == password
-
-	if PassInitialized {
-		runPass("", "rm", "-rf", name)
-	}
+func IsPresent() bool {
+	_, err := exec.LookPath("pass")
+	fmt.Println(err)
```
silvin-lubecki: fmt.Println(err) ?

pass/pass_linux.go
```diff
@@ -95,12 +79,38 @@ func runPass(stdinContent string, args ...string) (string, error) {
...
+	password := "pass is initialized"
+	name := path.Join(PASS_FOLDER, "docker-pass-initialized-check")
+
+	_, err := runPass(password, "insert", "-f", "-m", name)
+	if err != nil {
+		return false
+	}
+
+	stored, err := runPass("", "show", name)
+	passInitialized = err == nil && stored == password
```
silvin-lubecki: The error is silent here, maybe you should forward it to the caller?

pass/pass_linux.go
```diff
@@ -95,12 +79,38 @@ func runPass(stdinContent string, args ...string) (string, error) {
 	return string(result), nil
 }
 
+func isInitialized() bool {
+	if passInitialized {
```
silvin-lubecki: If an error occurs during the first pass of `isInitialized`, then `passInitialized` won't be true, and then we will go through `isInitialized` again, and maybe again etc... I think you should return an error if it fails and let the caller prevent from calling `isInitialized` again.

pass/pass_linux.go
thaJeztah: Should we ignore `permission denied` errors etc? (Was thinking if we should print any error other than`ErrNotFound`).

Also wondering if there's a need to cache the result, but possibly better left to the software using this

euank: I think we shouldn't bother providing this 2-line helper function. The caller can decide to check if `pass` is installed themselves just fine.

pass/pass_linux.go
n4ss: @silvin-lubecki and in this case you just disable this credentials storing method?

pass/pass_linux.go
n4ss: @euank I don't agree with that, a helper is meant to be pluggable and should remove the need for the caller to be aware of implementation details.

pass/pass_linux.go
n4ss: agree, would be cool to return a bool + err .

pass/pass_linux.go
silvin-lubecki: Why not? If there is an error, it means this credentials storing method is not available no? We could add a retry pattern here, but I don't think it would really help...

PR #108, opened 2018-06-21
pass: trim pass show output

### LINKED ISSUE #107: "pass" helper doesn't work with pass versions after 1.7.0
password store, as of a [recent commit](https://git.zx2c4.com/password-store/commit/?id=8446a40fbca8b092d28952edba163ed7df0be2a2), switched to buffering password output and including a newline on the end via using "echo" to then print that buffered data.

This breaks the pass implementation, causing the init function to never consider it initialized.

euank:
As of 8446a40, pass show will include a newline when showing a password.
This causes the pass helper here to reliably fail to initialize since a
password doesn't round-trip.

Before making this change, the pass test would fail if the installed
password-store version was v1.7.1+, and after this change it passes
again.

Fixes #107

Signed-off-by: Euan Kemp <euank@euank.com>


n4ss:
LGTM!

vdemeester:
Rebased against master to get CI green :angel:

raghuvanshy:
Have been facing the same issue. Has the patch been made available as an update?

teohhanhui:
@raghuvanshy It's in v0.6.1

PR #120, opened 2018-08-02
Fix typo and spelling error

efine:
This just fixes a typo and a spelling error in the comments.

PR #92, opened 2017-09-01
correct spelling mistake

EdwardBetts:

PR #124, opened 2018-09-22
pass: changed the way for checking if password-store is initalized

eyJhb:
Changed the way that `docker-credential-helpers` check if `password-store` is initialized.
The current way inserts and removes a password, each time it is called, to make sure that pass is initialized.

This creates a lot of commits if the password-store is made as a git repository (which is quite awful).
The following is how many commits has been made, over a span of ~2 months...

```
$ git log | grep 'docker-pass-initialized-check' | wc -l
1430
```

I have changed the way to just do a `pass ls`, as it will fail if pass is not initialized.
If it should fail, it will fail at a read of pass or insert of pass anyways, and therefore this solution is much more elegant.

### DIFF
--- a/pass/pass_linux.go
+++ b/pass/pass_linux.go
@@ -46,25 +46,10 @@ func (p Pass) checkInitialized() error {
 	if passInitialized {
 		return nil
 	}
-	// In principle, we could just run `pass init`. However, pass has a bug
-	// where if gpg fails, it doesn't always exit 1. Additionally, pass
-	// uses gpg2, but gpg is the default, which may be confusing. So let's
-	// just explictily check that pass actually can store and retreive a
-	// password.
-	password := "pass is initialized"
-	name := path.Join(getPassDir(), "docker-pass-initialized-check")
-
-	_, err := p.runPassHelper(password, "insert", "-f", "-m", name)
+	// We just run a `pass ls`, if it fails then pass is not initialized.
+	_, err := p.runPassHelper("", "ls")
 	if err != nil {
 		return fmt.Errorf("error initializing pass: %v", err)
-	}
-
-	stored, err := p.runPassHelper("", "show", name)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


eyJhb:
This also fixes #120 #92, since those lines has been removed.

pass/pass_linux.go
```diff
@@ -46,26 +46,11 @@ func (p Pass) checkInitialized() error {
...
-	// just explictily check that pass actually can store and retreive a
-	// password.
-	password := "pass is initialized"
-	name := path.Join(getPassDir(), "docker-pass-initialized-check")
-
-	_, err := p.runPassHelper(password, "insert", "-f", "-m", name)
+	// We just run a `pass ls`, if it fails then pass is not initialized.
+	_, err := p.runPassHelper("", "ls")
 	if err != nil {
 		return fmt.Errorf("error initializing pass: %v", err)
```
daltonmatos: Hello @eyJhb,

What do you think about changing the error message to something on these lines?

`password-store not initialized: %v`. 

Since we are not trying to intialize the password-store, this message could confuse users.

eyJhb:
Just a little ping, @vdemeester @n4ss if you get the time to review and merge this.
Would really like some feedback, and it seems like this can be a somewhat inactive repo at times. :)

eyJhb:
Great! Do not hesitate if you have any questions or I need to do anything, so we can get this merged and close the two PRs and the one issue :)

nassycheezy:
LGTM!

javabrett:
I've been seeing my `pass` db for Docker regularly corrupted by insertion and non-deletion of an incomplete `docker-pass-initialized-check` key, so it would be positive to see this change released promptly from `master`.

javabrett:
OK here's another essential tip if you find your `pass` repo constantly corrupted by incomplete `docker-pass-initialized-check`:

```
export GPG_TTY=$(tty)
```

... so that `gpg` properly prompts for key passphrase, assuming you have one.

PR #130, opened 2018-11-30
Add Version to the helper interface.

calavera:
So each helper can advertise their own version.


guillaumerose:
Can you rebase ? The CI is fixed now.

PR #139, opened 2019-04-03
make docker-credential-wincred work like docker-credential-osxkeychain

ekcasey:
* fetch credentials for server with matching hostname if scheme, path, or port are not provided
* if the credential request includes specific scheme, path, or port that does not match entry, don't return
* extract url helpers into a package

Signed-off-by: Danny Joyce <djoyce@pivotal.io>
Signed-off-by: Emily Casey <ecasey@pivotal.io>


ekcasey:
This fixes issues fetching docker credentials in libraries like https://github.com/google/go-containerregistry that resolve Docker Hub registry url differently (https://index.docker.io vs https://index.docker.io/v1)

ekcasey:
@vdemeester It looks like the Travis build is broken for reasons unrelated to our changes. If that is not a correct diagnosis, let @djoyahoy and I know so we fix it :)

guillaumerose:
CI is now fixed. Can you rebase ? Thanks

ekcasey:
@guillaumerose done!

osxkeychain/osxkeychain_darwin.go
```diff
@@ -10,9 +10,8 @@ package osxkeychain
 import "C"
 import (
 	"errors"
-	"net/url"
+	"github.com/docker/docker-credential-helpers/registryurl"
```
silvin-lubecki: nit picking: Can you re-order the imports? `goimports` should do that for you 😺

wincred/wincred_windows.go
```diff
@@ -2,6 +2,8 @@ package wincred
 
 import (
 	"bytes"
+	"github.com/docker/docker-credential-helpers/registryurl"
```
silvin-lubecki: nit: please re-organize the imports here too

wincred/wincred_windows.go
```diff
@@ -51,6 +61,72 @@ func (h Wincred) Get(serverURL string) (string, string, error) {
 	return "", "", credentials.NewErrCredentialsNotFound()
 }
 
+func getTarget(serverURL string) (string, error) {
+	s, err := registryurl.Parse(serverURL)
+	if err != nil {
+		return serverURL, nil
```
silvin-lubecki: nit: return empty string here instead?

wincred/wincred_windows.go
```diff
@@ -51,6 +61,72 @@ func (h Wincred) Get(serverURL string) (string, string, error) {
...
+	}
+
+	creds, err := winc.List()
+	if err != nil {
+		return "", err
+	}
+
+	targets := make([]string, 0)
```
silvin-lubecki: nit: `var targets []string`

wincred/wincred_windows.go
```diff
@@ -51,6 +61,72 @@ func (h Wincred) Get(serverURL string) (string, string, error) {
...
+	for i := range creds {
+		attrs := creds[i].Attributes
+		for _, attr := range attrs {
+			if strings.Compare(attr.Keyword, "label") == 0 &&
```
silvin-lubecki: From the documentation: https://golang.org/pkg/strings/#Compare
> Compare is included only for symmetry with package bytes. It is usually clearer and always faster to use the built-in string comparison operators ==, <, >, and so on.

So I guess `if attr.Keyword == "label"` is simpler 😸

wincred/wincred_windows.go
```diff
@@ -51,6 +61,72 @@ func (h Wincred) Get(serverURL string) (string, string, error) {
...
+				bytes.Compare(attr.Value, []byte(credentials.CredsLabel)) == 0 {
```
silvin-lubecki: You should use [bytes.Equal](https://golang.org/pkg/bytes/#Equal) instead.
What is the real type behind `attr.Value`? Is it an arbitrary byte slice? Or always a string encoded into bytes? My point is: even if the result is the same, I think it's more clearer to cast `attr.Value` as string instead.

wincred/wincred_windows.go
ekcasey: We returned the serverURL here b/c before we made our changes there was never a requirement that the url had to parse successfully (you could still fetch an exact match from the cred store). Happy to change this if you want to make this a requirement.

silvin-lubecki: Oh right read it too fast, you return a nil error. That's fine then 👍

wincred/wincred_windows.go
ekcasey: This piece of code existed before our PR so weren't sure if `attr.Value` needed to be a byte slice for a specific reason, happy to make the changes though.

wincred/wincred_windows.go
```diff
@@ -51,6 +61,72 @@ func (h Wincred) Get(serverURL string) (string, string, error) {
...
+				targets = append(targets, creds[i].TargetName)
+			}
+		}
+	}
+
+	if target, found := findMatch(s, targets, exactMatch); found {
```
silvin-lubecki: Is this really needed? Can't we just match with approximation? What is the targets length ?

ekcasey: We assumed that if there were both an exact and an approximate match in the cred store, the exact match would be preferable. We can take the exact match out if you disagree.

silvin-lubecki: Sounds good 😄 Ok then 👍

ekcasey:
Thanks everybody!

PR #142, opened 2019-04-17
Fix Travis configuration, update XCode version, Golang version and set to minimum macOS version to 10.11.

ulrich:
Using the new public key for RVM, update to golang 1.9.1 version…

Signed-off-by: Ulrich VACHON <ulrich.vachon@docker.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,19 +3,14 @@
   sudo: required
   language: go
   dist: trusty
-  osx_image: xcode9
+  osx_image: xcode10.1
   os:
     - linux
     - osx
   notifications:
     email: false
   go:
-    - 1.8.1
-  before_install:
-    # work-around for issue https://github.com/travis-ci/travis-ci/issues/6307
-    # might not be necessary in the future
-    - command curl -sSL https://rvm.io/mpapis.asc | gpg --import -
-    - rvm get stable
+    - 1.12.x
   addons:
     apt:
       packages:
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ VERSION := $(shell grep 'const Version' credentials/version.go | awk -F'"' '{ pr
 all: test
 
 deps:
-	go get -u github.com/golang/lint/golint
+	go get -u golang.org/x/lint/golint
 
 clean:
 	rm -rf bin
--- a/credentials/version.go
+++ b/credentials/version.go
@@ -1,4 +1,4 @@
 package credentials
 
 // Version holds a string describing the current version
-const Version = "0.6.0"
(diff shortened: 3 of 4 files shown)


guillaumerose:
Can we bump to go1.12 ?

ulrich:
> Can we bump to go1.12 ?

I wanted to move in small steps but it is in my projects.

ulrich:
Bump to go1.12

.travis.yml
```diff
@@ -3,18 +3,18 @@
...
   notifications:
     email: false
   go:
-    - 1.8.1
+    - 1.12.1
   before_install:
     # work-around for issue https://github.com/travis-ci/travis-ci/issues/6307
     # might not be necessary in the future
-    - command curl -sSL https://rvm.io/mpapis.asc | gpg --import -
+    - command gpg --verbose --keyserver hkp://fks.pgpkeys.eu --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
```
thaJeztah: did we check if the workaround is still needed? I see https://github.com/travis-ci/travis-ci/issues/6307 was closed, so perhaps updating is no longer needed?

.travis.yml
```diff
@@ -3,18 +3,18 @@
...
-  osx_image: xcode9
+  osx_image: xcode10.1
   os:
     - linux
     - osx
   notifications:
     email: false
   go:
-    - 1.8.1
+    - 1.12.1
```
thaJeztah: this should be 1.12.4 now, but perhaps better to use `"1.12.x"` (with the quotes as well)

CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ This changelog tracks the releases of docker-credential-helpers.
 This project includes different binaries per platform.
 The platform released is identified after the tag name.
 
+## v0.6.1 (Go client, Mac OS X)
```
thaJeztah: could you move the changelog update and version bump to a separate PR?

I think there's more changes since the last release, and it's better to do it separate to review that properly

.travis.yml
ulrich: Indeed, this workaround is no more necessary.

.travis.yml
ulrich: Using 1.12.x instead of.

CHANGELOG.md
ulrich: OK.

ulrich:
@thaJeztah thank you for your feedbacks, I did the required modifications.

Issue #146, opened 2019-05-04
pass variant macos support


glensc:
looks like it's assumed that [`pass`](https://www.passwordstore.org/) is to be used on Linux only.

however `pass` written in bash, a frontend to `gpg`, and therefore is portable and works fine on macos, there's even brew [formula](https://github.com/Homebrew/homebrew-core/blob/9b37cb405386ed11af4cb68130d56ceb5335d413/Formula/pass.rb) available.

however build system assumes it's for Linux:

```
==> make pass
Last 15 lines from /Users/glen/Library/Logs/Homebrew/docker-credential-helper/03.make:
2019-05-04 18:16:10 +0300

make
pass

mkdir -p bin
go build -o bin/docker-credential-pass pass/cmd/main_linux.go
pass/cmd/main_linux.go:5:2: build constraints exclude all Go files in /private/tmp/docker-credential-helper-20190504-69257-nmhjjp/docker-credential-helpers-0.6.2/src/github.com/docker/docker-credential-helpers/pass
make: *** [pass] Error 1

```

in source code there's only Linux source:
- https://github.com/docker/docker-credential-helpers/tree/v0.6.2/pass

should be trivial to port to other platforms (macOS i'm interested now)

glensc:
note, this builds just fine:

```
==> cp pass/cmd/main_linux.go pass/cmd/main_darwin.go
==> cp pass/pass_linux.go pass/pass_darwin.go
==> go build -o bin/docker-credential-pass pass/cmd/main_darwin.go
```

hutson:
Looks like @ellsclytn filed a pull request #169 to allow `pass` to build on MacOS.

I'm looking forward to this change making it easier for Homebrew/Linuxbrew users to use the Docker Credentials Helper `pass` package across platforms for greater credential security and consistency. (https://github.com/Homebrew/homebrew-core/issues/39517)

PR #169, opened 2019-12-06
Allow pass helper to be built for macOS

ellsclytn:
Pass is described as "The Standard Unix Password Manager", and so it can
easily run on more than just Linux. Namely, it's supported on macOS. The
CLI is identical to the Linux build, which means the Linux helper code
for Pass is fully applicable toward the macOS build - a couple of
renames being the only needed thing, purely for semantic correctness.

Resolves https://github.com/docker/docker-credential-helpers/issues/146

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -109,6 +109,7 @@ RUN --mount=type=bind,target=. \
 EOT
 
 FROM base AS build-darwin
+ARG TARGETOS
 ARG TARGETARCH
 ARG TARGETVARIANT
 RUN --mount=type=bind,target=. \
@@ -121,6 +122,9 @@ RUN --mount=type=bind,target=. \
   xx-go install std
   xx-go build -ldflags "$(cat /tmp/.ldflags)" -o /out/docker-credential-osxkeychain-${TARGETARCH}${TARGETVARIANT} ./osxkeychain/cmd/
   xx-verify /out/docker-credential-osxkeychain-${TARGETARCH}${TARGETVARIANT}
+
+  xx-go build -ldflags "$(cat /tmp/.ldflags)" -o /out/docker-credential-pass-${TARGETOS}-${TARGETARCH}${TARGETVARIANT} ./pass/cmd/
+  xx-verify /out/docker-credential-pass-${TARGETOS}-${TARGETARCH}${TARGETVARIANT}
 EOT
 
 FROM base AS build-windows
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,7 @@ linuxrelease:
 osxrelease:
 	mkdir -p release
 	cd bin && tar cvfz ../release/docker-credential-osxkeychain-v$(VERSION)-amd64.tar.gz docker-credential-osxkeychain
+	cd bin && tar cvfz ../release/docker-credential-pass-v$(VERSION)-darwin-amd64.tar.gz docker-credential-pass
 
 winrelease:
 	mkdir -p release
--- a/pass/pass.go
+++ b/pass/pass.go
@@ -22,7 +22,7 @@ import (
 // PASS_FOLDER contains the directory where credentials are stored
 const PASS_FOLDER = "docker-credential-helpers" //nolint:revive
 
-// Pass handles secrets using Linux secret-service as a store.
+// Pass handles secrets using pass as a store.
 type Pass struct{}
 
 // Ideally these would be stored as members of Pass, but since all of Pass's


pass/pass.go
```diff
@@ -21,7 +21,7 @@ import (
 
 const PASS_FOLDER = "docker-credential-helpers"
 
-// Pass handles secrets using Linux secret-service as a store.
+// Pass handles secrets using secret-service as a store.
```
thaJeztah, line 24: Actually wondering here if the description was correct for the `pass` credentials helper, or if this was a copy/paste error from the sercretservice helper? ("using secret-service as a store")

ellsclytn: Fair point. I admittedly had no idea what `secret-service` was when I was working on this, so I decided to take the lazy way out and leave most of the comment untouched.

How's "Pass handles secrets using GPG encrypted files as a store" sound?

Or maybe the comment is entirely useless, I'm not really sure.

dolmen: 👍 for the comment change.

Note that `pass` is already described in the [package overview](https://pkg.go.dev/github.com/docker/docker-credential-helpers/pass), so the comment on that symbol is not that important.

thaJeztah:
Most of this PR was already merged as part of https://github.com/docker/docker-credential-helpers/pull/187 - I did a quick rebase to see what's remaining in this PR 😅

crazy-max:
Would need to also update `build-darwin` stage in `Dockerfile`

thaJeztah:
Yup, just realised, and just pushed before you commented 😂

Dockerfile
```diff
@@ -71,6 +71,9 @@ RUN --mount=type=bind,target=. \
   xx-go install std
   xx-go build -ldflags "$(cat /tmp/.ldflags)" -o /out/docker-credential-osxkeychain-${TARGETARCH}${TARGETVARIANT} ./osxkeychain/cmd/main_darwin.go
   xx-verify /out/docker-credential-osxkeychain-${TARGETARCH}${TARGETVARIANT}
+
+  xx-go build -ldflags "$(cat /tmp/.ldflags)" -o /out/docker-credential-pass-${TARGETARCH}${TARGETVARIANT} ./pass/cmd/main.go
+  xx-verify /out/docker-credential-pass-${TARGETARCH}${TARGETVARIANT}
```
crazy-max, line 76: should add also `${TARGETOS}` now that is not only linux.

thaJeztah: Ah, good one; wasn't sure if we collected them all in the same spot.

Let me have a look (we probably still need to look if we want to enable this by default and/or if that could mess up release pipelines elsewhere)

Dockerfile
```diff
@@ -71,6 +71,9 @@ RUN --mount=type=bind,target=. \
   xx-go install std
   xx-go build -ldflags "$(cat /tmp/.ldflags)" -o /out/docker-credential-osxkeychain-${TARGETARCH}${TARGETVARIANT} ./osxkeychain/cmd/main_darwin.go
   xx-verify /out/docker-credential-osxkeychain-${TARGETARCH}${TARGETVARIANT}
+
+  xx-go build -ldflags "$(cat /tmp/.ldflags)" -o /out/docker-credential-pass-${TARGETOS}-${TARGETARCH}${TARGETVARIANT} ./pass/cmd/main.go
+  xx-verify /out/docker-credential-pass-${TARGETOS}-${TARGETARCH}${TARGETVARIANT}
```
crazy-max, line 76: I think you need to define `ARG TARGETOS` in this stage too.

Makefile
```diff
@@ -41,6 +41,7 @@ linuxrelease:
 osxrelease:
 	mkdir -p release
 	cd bin && tar cvfz ../release/docker-credential-osxkeychain-v$(VERSION)-amd64.tar.gz docker-credential-osxkeychain
+	cd bin && tar cvfz ../release/docker-credential-pass-v$(VERSION)-osx-amd64.tar.gz docker-credential-pass
```
thaJeztah, line 44: This probably doesn't match either 🤔 (I see `osx`)

crazy-max: hum indeed yes, I think `darwin` is more aligned with what we ship in other projects right?

Dockerfile
thaJeztah: 🤦 that's what you get when you try to do it too fast 😂

crazy-max (approved):

thaJeztah:
Rebased to get a fresh run of CI with the latest changes; still green, so let me get this one in 👍

PR #171, opened 2020-01-22
Enabled ppc64le under "arch" instead of specifying under "os"

vrushaliinamdar:
Enabled ppc64le under "arch" instead of specifying under "os"
arch:
    - ppc64le


SuperSandro2000:
> Amending updates the existing PR. You DO NOT need to open a new one.

PR #191, opened 2020-11-24
List wraps the error from secretservice_linux.c.

QiWang19:
List wraps the error from secretservice_linux.c so the List can give the caller more useful error message for diagnosing.

Signed-off-by: Qi Wang <qiwan@redhat.com>

### DIFF
--- a/secretservice/secretservice_linux.go
+++ b/secretservice/secretservice_linux.go
@@ -96,7 +96,8 @@ func (h Secretservice) List() (map[string]string, error) {
 	defer C.freeListData(&acctsC, listLenC)
 	if err != nil {
 		defer C.g_error_free(err)
-		return nil, errors.New("Error from list function in secretservice_linux.c likely due to error in secretservice library")
+		errMsg := (*C.char)(unsafe.Pointer(err.message))
+		return errors.New(C.GoString(errMsg))
 	}
 
 	resp := make(map[string]string)


secretservice/secretservice_linux.go
```diff
@@ -96,7 +96,8 @@ func (h Secretservice) List() (map[string]string, error) {
 	defer C.freeListData(&acctsC, listLenC)
 	if err != nil {
 		defer C.g_error_free(err)
-		return nil, errors.New("Error from list function in secretservice_linux.c likely due to error in secretservice library")
+		errMsg := (*C.char)(unsafe.Pointer(err.message))
+		return nil, errors.New("Error from list function in secretservice_linux.c likely due to error in secretservice library" + ": " + C.GoString(errMsg))
```
ebriney, line 100: just do `return errors.New(C.GoString(errMsg))` like https://github.com/docker/docker-credential-helpers/blob/4b9fe97b79d3d6c1bd77b91bc18b89efd21a7ace/secretservice/secretservice_linux.go#L37 and https://github.com/docker/docker-credential-helpers/blob/4b9fe97b79d3d6c1bd77b91bc18b89efd21a7ace/secretservice/secretservice_linux.go#L53

ebriney (changes requested):

secretservice/secretservice_linux.go
QiWang19: @ebriney Fixed, thanks!

ebriney (approved):
LGTM

PR #168, opened 2019-12-04
Add go.mod, and vendor wincred v1.1.0

thaJeztah:
Committing the result of

```console
go mod init
go mod vendor
```

However, trying to write code to _validate_ the vendored files, I get some weird results; when running

```console
go mod tidy
```

go mod adds various test-dependencies to `go.mod`, and I cannot find where they're referenced (not even in the vendored dependency ?); diff below

<details>

```patch
diff --git a/go.mod b/go.mod
index d24aa62..ec58726 100644
--- a/go.mod
+++ b/go.mod
@@ -2,4 +2,7 @@ module github.com/docker/docker-credential-helpers

 go 1.13

-require github.com/danieljoos/wincred v1.0.2
+require (
+       github.com/danieljoos/wincred v1.0.2
+       github.com/stretchr/testify v1.4.0 // indirect
+)
diff --git a/go.sum b/go.sum
index c531bc4..9724c6d 100644
--- a/go.sum
+++ b/go.sum
@@ -1,2 +1,14 @@
 github.com/danieljoos/wincred v1.0.2 h1:zf4bhty2iLuwgjgpraD2E9UbvO+fe54XXGJbOwe23fU=
 github.com/danieljoos/wincred v1.0.2/go.mod h1:SnuYRW9lp1oJrZX/dXJqr0cPK5gYXqx3EJbmjhLdK9U=
+github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8=
+github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
+github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
+github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
+github.com/stretchr/objx v0.1.0 h1:4G4v2dO3VZwixGIRoQ5Lfboy6nUhCyYzaqnIAPPhYs4=
+github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
[...]

PR #195, opened 2021-01-21
Use golang.org/x/sys/execabs

tiborvass:
Depends on #168


client/command.go
```diff
@@ -4,7 +4,8 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"os/exec"
+
+	exec "golang.org/x/sys/execabs"
```
thaJeztah, line 8: oh, nice; I missed that they included it there as well, I thought it was only in an `internal/` package (https://github.com/golang/go/commit/07e3195293ec510171d7d43ec8ac2bcb9cf00df#diff-25560022006dcc0a0bf07fe3291119c24ca8b12d0f29d3ad01f7c8a55265af58R34-R43)

thaJeztah (approved):
LGTM

PR #201, opened 2021-05-08
Set a better displaylabel for secretservice

WhyNotHugo:
Secretservice entries have a "label". This is intended to be a
human-readable description. It's actually called "Description" in UIs
like seahorse, and the listing of existing secrets shows this as a name
for each one.

The entries stored by the credential helper set this to simply the
repository URL. This is rather unfriendly, since entries like
"gitlab.com" and "index.docker.io/v1" show up. Mixed in with
entries from all other applications, it's hard to figure out what
application owns each entry.

This commit changes the label used when saving entries to be something
human-readable (this is the intent of the "label" field, btw). Because
of the naming scheme, this also results in all entries being shown
together by default (since UIs tend to sort lexicographically).

New entries will now be stores as:

    Docker credentials for $REGISTRY_URL

Note that items stored by the secret service have multiple fields inside
of them. One of those fields is called "label", and is used by the
helper to filter items from the secret service. This "label" field is
entirely unrelated to the items' label. The naming is most unfortunate.

### DIFF
--- a/secretservice/secretservice_linux.c
+++ b/secretservice/secretservice_linux.c
@@ -17,11 +17,11 @@ const SecretSchema *docker_get_schema(void)
 	return &docker_schema;
 }
 
-GError *add(char *label, char *server, char *username, char *secret) {
+GError *add(char *label, char *server, char *username, char *secret, char *displaylabel) {
 	GError *err = NULL;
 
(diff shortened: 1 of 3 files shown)


WhyNotHugo:
This is what entries look on a Seahorse (a UI to inspect one's secretservice storage).

Without this patch:

before


With this patch:

after

WhyNotHugo:
Any feedback for this? I don't think the CI failure is related.

PR #211, opened 2021-09-08
Fix return values in List() in secretservice_linux.go

alakesh:
In case of error, return nil for expected map along with error

Signed-off-by: Alakesh Haloi <alakeshh@amazon.com>


mat007 (approved):
Thanks!

PR #209, opened 2021-08-30
Switch to GitHub Actions

crazy-max:
Follow-up #208 (cc @thaJeztah)

This PR allows to build credential helpers on GitHub Actions using cross-compilation thanks to [tonistiigi/xx](https://github.com/tonistiigi/xx/) (cc @tonistiigi).

* Containerized environment, no need windows or macos runners
* Switch to GolangCI Lint
* Deb packaging refactored (containerized too)

Run example: https://github.com/crazy-max/docker-credential-helpers/actions/runs/1181534227

@ebriney Is Jenkinsfile still used?

Signed-off-by: CrazyMax <crazy-max@users.noreply.github.com>

PR #219, opened 2022-05-08
Dockerfile for cross compilation

crazy-max:
Follow-up #209 to split a bit what was done.

This one adds a Dockerfile to ease cross compilation and a simple GHA workflow that will just build the Dockerfile. Jenkinsfile has been removed as it's not used anymore. In another follow-up we can move the travis pipeline to GHA.

Build example: https://github.com/crazy-max/docker-credential-helpers/actions/runs/2288625907

cc @thaJeztah @ebriney


ebriney (approved):
Nice! LGTM

Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+  xx-go build -ldflags "$(cat /tmp/.ldflags)" -o /out/docker-credential-wincred-${TARGETARCH}${TARGETVARIANT}.exe ./wincred/cmd/main_windows.go
+  xx-verify /out/docker-credential-wincred-${TARGETARCH}${TARGETVARIANT}.exe
+EOT
+
+FROM build-$TARGETOS AS build
+
+FROM scratch AS binaries
+COPY --from=build /out /
+
+FROM binaries
```
thaJeztah, line 82: 👀 why the `FROM` here? 🤔

Dockerfile
```diff
@@ -0,0 +1,82 @@
+# syntax=docker/dockerfile:1
+
+ARG GO_VERSION=1.16.7
```
thaJeztah, line 3: (After this?) we should update to a current version of Go

Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+ARG XX_VERSION=1.1.2
+ARG OSXCROSS_VERSION=11.3-r7-alpine
+
+ARG PKG=github.com/docker/docker-credential-helpers
+
+# xx is a helper for cross-compilation
+FROM --platform=$BUILDPLATFORM tonistiigi/xx:${XX_VERSION} AS xx
+
+# osxcross contains the MacOSX cross toolchain for xx
+FROM crazymax/osxcross:${OSXCROSS_VERSION} AS osxcross
```
thaJeztah, line 13: This is built from https://github.com/crazy-max/osxcross ? Or from the upstream one? (I see neither repo tags releases, so was wondering what version it was built from)

thaJeztah (commented):
overall looks good (thanks!) some questions

Dockerfile
crazy-max: Yes just keeping minimal atm

Dockerfile
crazy-max: > This is built from https://github.com/crazy-max/osxcross ?

It's built from https://github.com/crazy-max/docker-osxcross

Dockerfile
crazy-max: To make it the default one when building. I could also just `FROM scratch` the previous one but I tend to prefer naming stages so the bake definition fits better.

crazy-max: Oh nevermind I can just remove it I think

thaJeztah (approved):
LGTM

Issue #220, opened 2022-07-27
'pass' based credential helper returns success when cred doesn't exist


nicks:
Repro steps:

```
echo "garbage" | go run ./pass/cmd/main.go get
```

Expected:
Exit code 1

Actual:
Exit code 0
{"ServerURL":"garbage","Username":"","Secret":""}

This is the root of cryptic errors like:

```
WARNING: could not erase credentials: error erasing credentials - err: exit status 1, out: `exit status 1: Error: docker-credential-helpers/[CRED_FOLDER_NAME] is not in the password store.` 
```

ericpromislow:
The thing I find strange is that calling the underlying function `pass show security` would write `Error: security is not in the password store` to stderr and exit with status 1. But if I remember looking at this code a while ago, it's walking the raw filesystem.

This is related to the other problem that you can't use `docker-credential-pass list` to determine if the docker password-store has even been initialized because it returns `{}`, even if `pass (1)` hasn't been installed.

PR #221, opened 2022-07-27
pass: return an error when a cred doesn't exist

nicks:
fixes https://github.com/docker/docker-credential-helpers/issues/220

### DIFF
--- a/pass/pass.go
+++ b/pass/pass.go
@@ -138,7 +138,7 @@ func (p Pass) Get(serverURL string) (string, string, error) {
 
 	if _, err := os.Stat(path.Join(getPassDir(), PASS_FOLDER, encoded)); err != nil {
 		if os.IsNotExist(err) {
-			return "", "", nil
+			return "", "", credentials.NewErrCredentialsNotFound()
 		}
 
 		return "", "", err
--- a/pass/pass_test.go
+++ b/pass/pass_test.go
@@ -54,13 +54,9 @@ func TestPassHelper(t *testing.T) {
 			t.Fatal(err)
 		}
 
-		username, _, err = helper.Get(server)
-		if err != nil {
-			t.Fatal(err)
-		}
-
-		if username != "" {
-			t.Fatalf("%s shouldn't exist any more", username)
+		_, _, err = helper.Get(server)
+		if !credentials.IsErrCredentialsNotFound(err) {
+			t.Fatalf("expected credentials not found, actual: %v", err)
 		}
 	}
 
@@ -71,5 +67,14 @@ func TestPassHelper(t *testing.T) {
 
 	if len(credsList) != 0 {
 		t.Fatal("didn't delete all creds?")
+	}
+}
+
+func TestMissingCred(t *testing.T) {
+	helper := Pass{}
+
+	_, _, err := helper.Get("garbage")
+	if !credentials.IsErrCredentialsNotFound(err) {
+		t.Fatalf("expected credentials not found, actual: %v", err)
 	}
 }


pass/pass.go
```diff
@@ -138,7 +138,7 @@ func (p Pass) Get(serverURL string) (string, string, error) {
 
 	if _, err := os.Stat(path.Join(getPassDir(), PASS_FOLDER, encoded)); err != nil {
 		if os.IsNotExist(err) {
-			return "", "", nil
+			return "", "", credentials.NewErrCredentialsNotFound()
 		}
 
 		return "", "", err
```
thaJeztah, line 144: Looks like this whole bock was explicitly added in https://github.com/docker/docker-credential-helpers/pull/84 (linked to https://github.com/docker/cli/pull/451).

I don't recall the specifics though; this may have been related to _other_ binaries named `pass` on Linux, which would get called, so the check for that was to see if a credentials store was initialised, and otherwise fall back to regular credentials storage 🤔

nicks: The PR's description is:

> It turns out the cred helpers protocol is to return "" and not an error when a credential isn't present, so let's do that.

Is it possible the protocol changed since then?

I have a hard time reconciling that with the test. I didn't write this new test; I copied it from the other helpers:
https://github.com/docker/docker-credential-helpers/blob/master/wincred/wincred_windows_test.go#L238
https://github.com/docker/docker-credential-helpers/blob/master/osxkeychain/osxkeychain_darwin_test.go#L207
which confirm that this is supposed to return an error.

nicks: maybe the API contract changed to return a standardized error instead of an implementation-specific one?

I also considered refactoring the tests into a "API invariants" test suite that we share across all the helpers. happy to do that in this PR (or in a follow-up one)

thaJeztah: I'm not aware of changes in the API, but must really admit that at the time I wasn't closely involved in development of the "credential-helpers".
 

Thinking out loud here; as that patch was opened related to https://github.com/docker/cli/pull/451, could it be that (as, unlike macOS and Windows) Linux had multiple possible helpers, that it was implemented as a "hack" to;

- check if the `pass` backend can be supported (don't error on "missing")
- and _only_ fall back to the second alternative (`secretservice` helper) otherwise
- and as last resort, fallback to "store inside the config-file"

thaJeztah:
Hm... GitHub actions having issues? I see it was triggered, but didn't complete 😞.

Let me try the "good'ol" close & reopen.

pass/pass.go
nicks: i think that's a reasonable hypothesis, but i don't think that's what's happening. The switching code is pretty straightforward. It calls Get() here:

https://github.com/docker/cli/blob/master/cli/config/credentials/native_store.go#L118

and the switch between pass and secretservice is here:

https://github.com/docker/cli/blob/master/cli/config/credentials/default_store_linux.go#L8

And as far as I can tell, `pass/pass.go` NEVER implemented this protocol correctly. Here's the first commit in August 2017:
https://github.com/docker/docker-credential-helpers/commit/1ab10377076c7cd77d4ee210dc101f04486e1fbe
even though the macos credential helper returned this error in at least February 2016:
https://github.com/docker/docker-credential-helpers/commit/4b716da07041687901f4ee4ab37134753965b7f9

thaJeztah (approved):
Thanks for also doing a bit of digging! Yes, it appears that this error is not needed for the selection, so I _think_ this would be ok to go in (famous last words).

We should probably do a bit of cleaning up on the cli side as well (was thinking we should reverse the lookup to first check if we have a helper installed, and only then try if the `pass` binary is found)

Either way; that's in another repository.

We're planning to do a `-beta` release to test the latest changes in this repository, so let's get this one in, so that we can give it a try 👍 

LGTM

thaJeztah:
(Asking @crazy-max to have a last look before we merge; just in case)

crazy-max (approved):
Yes as @thaJeztah said, lookup should be the other way around to make sure `pass` folder exists first.

LGTM in the meantime.

PR #224, opened 2022-08-20
Test stage and ci jobs

crazy-max:
follow-up https://github.com/docker/docker-credential-helpers/pull/209

Tests run on specific runners with respect to the target OS and the results are merged and sent to Codecov. Containerized tests are only available for host platform atm (that's why a specific job has been added in the build workflow).

cc @thaJeztah

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,10 +11,71 @@ on:
 
 env:
   DESTDIR: ./bin
+  GO_VERSION: 1.16.7
 
 jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-20.04
+          - macOS-11
+          - windows-2022
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+      -
+        name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          cache: true
+      -
+        name: Install deps
+        if: ${{ matrix.os == 'ubuntu-20.04' }}
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y dbus-x11 gnome-keyring libsecret-1-dev pass
+      -
+        name: Test
+        run: |
+          go test -short -v -coverprofile=./coverage.txt -covermode=atomic ./...
+          go tool cover -func=./coverage.txt
+        shell: bash
+      -
+        name: Upload coverage
+        uses: codecov/codecov-action@v3
+        with:
+          file: ./coverage.txt
+
+  test-sandboxed:
+    runs-on: ubuntu-20.04
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v3
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      -
+        name: Test
+        uses: docker/bake-action@v2
+        with:
+          targets: test
+          set: |
+            *.cache-from=type=gha,scope=test
+            *.cache-to=type=gha,scope=test,mode=max
+      -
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -11,8 +11,69 @@ on:
...
+          go tool cover -func=./coverage.txt
+        shell: bash
+      -
+        name: Upload coverage
+        uses: codecov/codecov-action@v3
+        with:
+          file: ./coverage.txt
+
+  test-ctn:
+    runs-on: ubuntu-latest
```
thaJeztah, line 55: Should this be `ubuntu-20.04` as well?

.github/workflows/build.yml
```diff
@@ -11,8 +11,69 @@ on:
...
+            *.cache-from=type=gha,scope=test
+            *.cache-to=type=gha,scope=test,mode=max
+      -
+        name: Upload coverage
+        uses: codecov/codecov-action@v3
+        with:
+          file: ${{ env.DESTDIR }}//coverage.txt
+
   build:
     runs-on: ubuntu-latest
```
thaJeztah, line 78: Same for this one

README.md
```diff
@@ -6,9 +12,18 @@ docker-credential-helpers is a suite of programs to use native stores to keep Do
 
 Go to the [Releases](https://github.com/docker/docker-credential-helpers/releases) page and download the binary that works better for you. Put that binary in your `$PATH`, so Docker can find it.
 
-### Building from scratch
+## Building
+
+You can easily build the credential helpers using Docker:
```
thaJeztah, line 17: Perhaps remove "easily"

README.md
```diff
@@ -6,9 +12,18 @@ docker-credential-helpers is a suite of programs to use native stores to keep Do
...
+
+```shell
+# create builder
+$ docker buildx create --use
```
thaJeztah, line 21: Does this _require_ a container builder? If it's just single arch, it should work without, I think?

README.md
```diff
@@ -6,9 +12,18 @@ docker-credential-helpers is a suite of programs to use native stores to keep Do
...
+# build credential helpers and output to ./bin/build
+$ docker buildx bake "https://github.com/docker/docker-credential-helpers.git"
```
thaJeztah, line 23: Wondering if people expect the build to be from remote source (more likely they would clone the repo?)

Perhaps we should mention both options, otherwise someone would clone the repo, and then build from the remote source 😬

README.md
```diff
@@ -6,9 +12,18 @@ docker-credential-helpers is a suite of programs to use native stores to keep Do
...
+```
 
-The programs in this repository are written with the Go programming language. These instructions assume that you have previous knowledge about the language and you have it installed in your machine.
+Or if the toolchain is already installed on your machine:
 
 1 - Download the source and put it in your `$GOPATH` with `go get`.
```
thaJeztah, line 28: Looks like this section is in need of a rewrite as well (`go get` won't work anymore to clone the repository)

README.md
crazy-max: Yes required because we are building against multi-platforms (osxkeychain, wincred, pass)

README.md
crazy-max: Ah yes good point, will add also how to do it from local source

README.md
crazy-max: Yes if we bump to Go 1.19 this has to be changed

README.md
thaJeztah: ah, right, it's building all of them

what's the easiest way to build only one?

README.md
thaJeztah: I prepared https://github.com/docker/docker-credential-helpers/pull/226 for go 1.18 (but best to do after this one)

thaJeztah (approved):
LGTM

PR #225, opened 2022-08-20
registryurl: remove fallback code for go < 1.8

thaJeztah:
Go 1.8 has ben EOL for years, so we can drop this


registryurl/parse.go
```diff
@@ -35,3 +35,13 @@ func Parse(registryURL string) (*url.URL, error) {
 	u.RawQuery = ""
 	return u, nil
 }
+
+// GetHostname returns the hostname of the URL
+func GetHostname(u *url.URL) string {
```
thaJeztah, line 40: Actually, let me drop internal uses of this function

crazy-max (approved):
LGTM

PR #172, opened 2020-01-22
Enabled ppc64le as an architecture

vrushaliinamdar:
Enabled ppc64le under "arch" instead of specifying under "os"
arch:
    - ppc64le

PR #227, opened 2022-08-20
support linux/ppc64le and linux/s390x cross build

crazy-max:
closes #172


thaJeztah (approved):
LGTM

PR #229, opened 2022-08-20
Dockerfile: de-duplicate args

thaJeztah:
When extending a stage, the new stage is considered a
breakpoint, so

    FROM foo AS one
    ARG something

    FROM one AS two
    # ARG is present here as well

Signed-off-by: Sebastiaan van Stijn <github@gone.nl>


thaJeztah:
FWIW; mostly opening this out of curiosity if it works (I wasn't a huge fan of this feature, as it may be confusing 😉)

PR #231, opened 2022-08-20
lint gha job and remove travis

crazy-max:
use golangci-lint and adds ci validation

this was the last piece in travis so we can remove travis config. I also found the `pass` tests were not covered because gpg key and pass was not initialized. Now pass tests run both on ci and in the Dockerfile.


.github/workflows/build.yml
```diff
@@ -11,7 +11,7 @@ on:
 
 env:
   DESTDIR: ./bin
-  GO_VERSION: 1.16.7
+  GO_VERSION: 1.18.5
```
thaJeztah, line 14: Ah, whoops, probably my bad (forgot to update after rebasing)

Makefile
```diff
@@ -50,28 +46,13 @@ test:
...
-vet_linux:
-	go vet ./secretservice
-
 lint:
-	for p in `go list ./... | grep -v /vendor/`; do \
-		golint $$p ; \
-	done
-
-fmt:
-	gofmt -s -l `ls **/*.go | grep -v vendor`
```
thaJeztah, line 71: Maybe this one is still handy to have to do a quick `gofmt` locally

ci/before_script_linux.sh
```diff
@@ -1,22 +0,0 @@
-set -ex
```
thaJeztah, line 1: Do we want to keep the script, in case we need to re-generate the fixtures?

thaJeztah (commented):
two comments, but otherwise looks good, thanks!

ci/before_script_linux.sh
crazy-max: Don't think it's necessary but sure

thaJeztah (approved):
LGTM, thanks!

PR #232, opened 2022-08-21
simplify "go build" commands

thaJeztah:
There's no need to specify the file to build, as we're building the "main" packages as a whole.


crazy-max (approved):

thaJeztah:
Oh! I think I see what may be wrong; does it work if you change the makefile to use; `./pass/cmd` (adding `./`) ?

PR #237, opened 2022-08-27
Release target with checksums

crazy-max:
Merge release targets in Makefile to a single one called `release`. Also create checksums and update build workflow to publish artifacts to GitHub Release on push tag.

Also adds extra commit to remove CHANGELOG as it's outdated and now part of GitHub Releases.


hack/release
```diff
@@ -0,0 +1,65 @@
...
+if ! command shasum --version > /dev/null 2>&1; then
+  echo >&2 "ERROR: shasum is required"
+  exit 1
+fi
+find ./${DESTDIR}/ -type f \( -iname "docker-credential-*" \) -print0 | sort -z | xargs -r0 shasum -a 256 -b | sed 's# .*/#  #' > ./${DESTDIR}/checksums.txt
+
+# verify
+(
+  cd ./${DESTDIR}
+  shasum -a 256 -U -c checksums.txt
```
thaJeztah, line 64: Looks like macOS / BSD and GNU are different here as well; the -U option is named -p on macOS;

On Linux:

```
 -U, --UNIVERSAL   read in Universal Newlines mode
                        produces same digest on Windows/Unix/Mac
```

On macOS

```
  -p, --portable    read in portable mode
                        produces same digest on Windows/Unix/Mac
```

So I guess we need to hack around that;

```suggestion
	if [ "$(uname -s)" = "Darwin" ]; then
		shasum -a 256 -p -c checksums.txt
	else
		shasum -a 256 -U -c checksums.txt
	fi
```

Dockerfile
```diff
@@ -85,54 +85,73 @@ EOT
...
+WORKDIR /work
+ARG TARGETOS
+ARG TARGETARCH
+ARG TARGETVARIANT
+RUN --mount=from=binaries \
+    --mount=type=bind,source=/tmp/.version,target=/tmp/.version,from=version <<EOT
+  set -e
+  mkdir /out
+  for f in *; do
+    cp "$f" "/out/${f%.*}-$(cat /tmp/.version).$TARGETOS-$TARGETARCH$TARGETVARIANT$([ "$TARGETOS" = "windows" ] && echo ".exe")"
```
thaJeztah, line 150: The multiple variables after each other makes it quite hard to read; perhaps use `{}` around them?

```suggestion
    cp "$f" "/out/${f%.*}-$(cat /tmp/.version).${TARGETOS}-${TARGETARCH}${TARGETVARIANT}$([ "$TARGETOS" = "windows" ] && echo ".exe")"
```

That said, there's a lot happening in that one line; I wonder if it would be more readable if some things are split. For example, the version would be the same for every file, so we could just capture that in a variable first. (Same for the file extension?)

```bash
  version="$(cat /tmp/.version)" \
  [ "$TARGETOS" = "windows" ] && file_ext=".exe"
  for f in *; do
    [ -n "${filename##*.}" ] && file_ext=".${filename##*.}"
    cp "$f" "/out/${f%.*}-${version}.${TARGETOS}-${TARGETARCH}${TARGETVARIANT}${file_ext}"
  done
```

It's a pity we need to mangle the name, as the original file already has the proper file extension 🤔. 

```bash
cp docker-credential-wincred.exe /out/docker-credential-wincred-v0.6.4-44-g0d1f2ec.m.windows-amd64.exe
```

Using a directory structure doesn't work for this? (Guess that's not an option if the goal is to publish to GitHub 😞 
Or.. would an alternative be too `.gz` them (keep the original name for the binary, but name the tar with version and platform?)

hack/release
crazy-max: > On macOS
> 
> ```
>   -p, --portable    read in portable mode
>                         produces same digest on Windows/Unix/Mac
> ```

I think you're using a pretty old version of perl utils. What's the output of `shasum -v`?

On my side on M1 (21.6.0) I don't have the `-p` flag available:

```console
$ uname -v
Darwin Kernel Version 21.6.0: Sat Jun 18 17:05:47 PDT 2022; root:xnu-8020.140.41~1/RELEASE_ARM64_T8101
$ shasum --help
Usage: shasum [OPTION]... [FILE]...
Print or check SHA checksums.
With no FILE, or when FILE is -, read standard input.

  -a, --algorithm   1 (default), 224, 256, 384, 512, 512224, 512256
  -b, --binary      read in binary mode
  -c, --check       read SHA sums from the FILEs and check them
      --tag         create a BSD-style checksum
  -t, --text        read in text mode (default)
  -U, --UNIVERSAL   read in Universal Newlines mode
                        produces same digest on Windows/Unix/Mac
  -0, --01          read in BITS mode
                        ASCII '0' interpreted as 0-bit,
                        ASCII '1' interpreted as 1-bit,
                        all other characters ignored

The following five options are useful only when verifying checksums:
      --ignore-missing  don't fail or report status for missing files
  -q, --quiet           don't print OK for each successfully verified file
  -s, --status          don't output anything, status code shows success
      --strict          exit non-zero for improperly formatted checksum lines
  -w, --warn            warn about improperly formatted checksum lines

  -h, --help        display this help and exit
  -v, --version     output version information and exit

When verifying SHA-512/224 or SHA-512/256 checksums, indicate the
algorithm explicitly using the -a option, e.g.

  shasum -a 512224 -c checksumfile

The sums are computed as described in FIPS PUB 180-4.  When checking,
the input should be a former output of this program.  The default
mode is to print a line with checksum, a character indicating type
(`*' for binary, ` ' for text, `U' for UNIVERSAL, `^' for BITS),
and name for each FILE.  The line starts with a `\' character if the
FILE name contains either newlines or backslashes, which are then
replaced by the two-character sequences `\n' and `\\' respectively.

Report shasum bugs to mshelor@cpan.org
```

Dockerfile
crazy-max: Change a bit to have smth more readable.

> Using a directory structure doesn't work for this? (Guess that's not an option if the goal is to publish to GitHub 😞
> Or.. would an alternative be too `.gz` them (keep the original name for the binary, but name the tar with version and platform?)

Don't think a tarball is necessary. Would also need to produce `.zip` for windows.

I think it's fine like this. We are not going to change that often anyway.

hack/release
```diff
@@ -0,0 +1,65 @@
...
+		signtool sign /fd SHA256 /a /f pfx /p ${PFXPASSWORD} /d Docker /du https://www.docker.com /t http://timestamp.verisign.com/scripts/timestamp.dll "$f"
+	done
+fi
+
+# checksums
+if ! command shasum --version > /dev/null 2>&1; then
+	echo >&2 "ERROR: shasum is required"
+	exit 1
+fi
+find ./${DESTDIR}/ -type f \( -iname "docker-credential-*" \) -print0 | sort -z | xargs -0 shasum -a 256 -b | sed 's# .*/#  #' > ./${DESTDIR}/checksums.txt
```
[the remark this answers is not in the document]
crazy-max: Good point, `shasum` is a perl util so `sha256sum` looks to be a better option. Also agree on simplification of release script for checksums.

credentials/version.go
```diff
@@ -2,7 +2,7 @@ package credentials
 
 var (
 	// Name is filled at linking time
-	Name = ""
+	Name = "unknown"
```
thaJeztah, line 5: Perhaps we should keep this empty, for cases where someone would "go install"; might be confusing?

thaJeztah (changes requested):
one last nit, but otherwise looks good

thaJeztah (approved):
LGTM

thaJeztah:
One difference with the previous releases is that previously we used a `tgz` to bundle the binary; advantage of that was that;

- after extracting, the binary has the right name
- and it looks like it has the right permissions (executable)

```
ls -la docker-credential-*
-rw-r--r-x 1 root root 2818053 Jun  7  2021 docker-credential-pass
-rw-r--r-- 1 root root 1781760 Aug 29 13:26 docker-credential-pass-v0.7.0-beta.1.linux-amd64
-rwxr-xr-x 1 root root 2495464 Jun  7  2021 docker-credential-secretservice
-rw-r--r-- 1 root root 1578552 Aug 29 13:32 docker-credential-secretservice-v0.7.0-beta.1.linux-amd64
```

I wasn't sure it the old ones were dynamically linked, but looks like `pass` static, but `secretservice` is dynamic (both previous release and new one);

```
ldd docker-credential-pass
	not a dynamic executable
ldd docker-credential-pass-v0.7.0-beta.1.linux-amd64
	not a dynamic executable
```

On debian:bullseye;

```
ldd docker-credential-secretservice
	linux-vdso.so.1 (0x00007ffd5b7da000)
	libsecret-1.so.0 => not found
	libgobject-2.0.so.0 => not found
	libglib-2.0.so.0 => not found
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f63c677e000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f63c65b9000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f63c67a4000)
ldd docker-credential-secretservice-v0.7.0-beta.1.linux-amd64
	linux-vdso.so.1 (0x00007ffcf137e000)
	libsecret-1.so.0 => not found
	libgio-2.0.so.0 => not found
	libgobject-2.0.so.0 => not found
	libglib-2.0.so.0 => not found
	libintl.so.8 => not found
	libc.musl-x86_64.so.1 => not found
```

On debian:buster;

```
ldd docker-credential-secretservice
	linux-vdso.so.1 (0x00007ffc289d4000)
	libsecret-1.so.0 => not found
	libgobject-2.0.so.0 => not found
	libglib-2.0.so.0 => not found
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f113bbbc000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f113b9fc000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f113bbe1000)

ldd docker-credential-secretservice-v0.7.0-beta.1.linux-amd64
	linux-vdso.so.1 (0x00007ffeff9f1000)
	libsecret-1.so.0 => not found
	libgio-2.0.so.0 => not found
	libgobject-2.0.so.0 => not found
	libglib-2.0.so.0 => not found
	libintl.so.8 => not found
	libc.musl-x86_64.so.1 => not found
```

There is a difference in both though on what it's linked against; what distro and libc does it require?

PR #239, opened 2022-08-30
Dockerfile: debian based build

crazy-max:
follow-up https://github.com/docker/docker-credential-helpers/pull/237#issuecomment-1230324172

using musl to build dynamically linked secretservice helper is not suitable so switch to debian based as it is for current released binaries.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,9 @@
 
 ARG GO_VERSION=1.18.5
 ARG XX_VERSION=1.1.2
-ARG OSXCROSS_VERSION=11.3-r7-alpine
+ARG OSXCROSS_VERSION=11.3-r7-debian
 ARG GOLANGCI_LINT_VERSION=v1.47.3
+ARG DEBIAN_FRONTEND=noninteractive
 
 ARG PACKAGE=github.com/docker/docker-credential-helpers
 
@@ -13,9 +14,10 @@ FROM --platform=$BUILDPLATFORM tonistiigi/xx:${XX_VERSION} AS xx
 # osxcross contains the MacOSX cross toolchain for xx
 FROM crazymax/osxcross:${OSXCROSS_VERSION} AS osxcross
 
-FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-alpine AS gobase
+FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-bullseye AS gobase
 COPY --from=xx / /
-RUN apk add --no-cache clang file git lld llvm make pkgconf rsync
+ARG DEBIAN_FRONTEND
+RUN apt-get update && apt-get install -y --no-install-recommends clang dpkg-dev file git lld llvm make pkg-config rsync
 ENV GOFLAGS="-mod=vendor"
 ENV CGO_ENABLED="1"
 WORKDIR /src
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -49,20 +51,25 @@ RUN --mount=type=bind,target=.,rw <<EOT
...
 
 FROM gobase AS base
 ARG TARGETPLATFORM
-RUN xx-apk add musl-dev gcc libsecret-dev pass
+ARG DEBIAN_FRONTEND
+RUN xx-apt-get install -y binutils gcc libc6-dev libgcc-10-dev libsecret-1-dev pkg-config
 
-FROM base AS test
-RUN xx-apk add gnome-keyring gpg-agent gnupg-gpgconf pass
+FROM golang:${GO_VERSION}-alpine AS test
```
crazy-max, line 68: unfortunately can't make sandboxed test work with debian base image :disappointed:

it seems `gpg` or `pass` does not behave like one alpine:

```
#15 0.721 ok    github.com/docker/docker-credential-helpers/credentials 0.002s  coverage: 58.5% of statements
#15 0.752 === RUN   TestPassHelper
#15 0.752     pass_test.go:20: exit status 1: gpg: WARNING: unsafe permissions on homedir '/root/.gnupg'
#15 0.752         gpg: 0xD523BD50DD70B0BA: There is no assurance this key belongs to the named user
#15 0.752         gpg: cannot open '/dev/tty': No such device or address
#15 0.752         Password encryption aborted.
#15 0.752
#15 0.752 --- FAIL: TestPassHelper (0.02s)
```

crazy-max: ah found the issue, on debian based we have to change the trust level command to:

```
echo '5\ny\n' | gpg --command-fd 0 --no-tty --edit-key 7D851EB72D73BDA0 trust
```

thaJeztah: For the permissions, we could probably also try to fix by setting the directory permissions when we create the directory (I recall I made that change when I was looking into things)

Dockerfile
```diff
@@ -13,9 +14,10 @@ FROM --platform=$BUILDPLATFORM tonistiigi/xx:${XX_VERSION} AS xx
 # osxcross contains the MacOSX cross toolchain for xx
 FROM crazymax/osxcross:${OSXCROSS_VERSION} AS osxcross
 
-FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-alpine AS gobase
+FROM --platform=$BUILDPLATFORM golang:${GO_VERSION}-bullseye AS gobase
```
thaJeztah, line 17: Do we know if `-buster` would work on both "bullseye" and "buster", or would that break? I recall the containerd maintainers had to downgrade their machines to ubuntu 18.04 to make them more compatible; https://github.com/containerd/containerd/compare/v1.6.7...v1.6.8

crazy-max: Yes I was also wondering what was our best option. I will try and let you know.

Dockerfile
thaJeztah: Yeah, was thinking; add a `mkdir` (and `chmod` if needed) here; https://github.com/docker/docker-credential-helpers/blob/2f246b8570b19c8ddb629149c7590a58e3715f06/Dockerfile#L70

Dockerfile
crazy-max: `ppc64le`, `s390x` and `armel` (arm/v6) cross pkgs are not available in buster :persevere:

thaJeztah: Alright; if we already used bullseye, then it's probably fine

thaJeztah (approved):
LGTM

PR #252, opened 2023-01-22
Fix build status badge being broken

okrc:
https://github.com/badges/shields/issues/8671

PR #253, opened 2023-01-22
Dockerfile: fix vendor validation

crazy-max:
follow-up https://github.com/docker/docker-credential-helpers/pull/252#issuecomment-1399598640

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,8 @@ FROM scratch AS vendor-update
 COPY --from=vendored /out /
 
 FROM vendored AS vendor-validate
-RUN --mount=type=bind,target=.,rw <<EOT
+RUN --mount=target=/context \
+    --mount=target=.,type=tmpfs <<EOT
   set -e
   rsync -a /context/. .
   git add -A


Dockerfile
```diff
@@ -38,7 +38,8 @@ FROM scratch AS vendor-update
 COPY --from=vendored /out /
 
 FROM vendored AS vendor-validate
-RUN --mount=type=bind,target=.,rw <<EOT
+RUN --mount=target=/context \
+    --mount=target=.,type=tmpfs <<EOT
```
thaJeztah, line 42: Wondering if we should change all the `.`'s with an absolute paths (I think it's `/src` in this case?). Currently I have to read "up" in the Dockerfile to understand what `.` refers to.

Do we know why a `tmpfs` is needed for this? Is this an issue with `overlayfs` not supporting some attributes, or what's the cause?

thaJeztah: Or is there no content at all (as there's no `COPY`) and that was the issue?

crazy-max: > Or is there no content at all (as there's no `COPY`) and that was the issue?

Yes that was the issue

thaJeztah (approved):
LGTM

Issue #162, opened 2019-07-25
docker-credential-pass does not work with username containing forward-slash


plakdawa:
On some docker registries the username is to be supplied as a chain of names each separated by forward-slash. For example [oracle cloud registry](https://docs.cloud.oracle.com/iaas/Content/Registry/Tasks/registrypushingimagesusingthedockercli.htm) uses the format tenancy/identityservice/username.

Instead of encrypting this string as base64 and storing it as a single key in the password store, docker-credential-pass stores it as chain of sub-keys. Even though `docker login` is successful, all subsequent interactions like `docker pull` fail.

sebageek:
I ran into the same problem - `docker-credential-pass` breaks when a `/` is in the username. I see these solutions:
* base64-encode the username, as it is already done with the server url
* replace `/` with a placeholder, e.g. `-SLASH-`
* do proper escaping (e.g. replace `/` with `-`, but also escape the escape character, `-` with `--`)

I have implemented base64-encoding of the username for me to get it working, but this breaks backwards compability. We could ignore backwards compability. We could also reencode everything that is non-base64 into base64. Still, it might not be the way to go. 

The proper-escaping solution is less likely to break backwards compability, but still might, depending upon the escape character.

Just replacing with `/` with something like `-SLASH-` would work as long as the replacement never appears in a docker username. `-SLASH-` is an unlikely candidate, but still not impossible.

Thoughts?

nasr18:
@plakdawa @sebageek any update on this? I'm facing this issue. Login success. But push/pull failing.

PR #267, opened 2023-05-10
Fix issue #162 (docker-credential-pass does not work with username containing forward-slash)

davcosentino:
Pull request to fix issue #162 (docker-credential-pass does not work with username containing forward-slash).

Based on [@sebageek](https://github.com/sebageek) [commit](https://github.com/sebageek/docker-credential-helpers/commit/56487ff15bb8f956e8989f72f4f8583cbb155db9).


davcosentino:
cc @thaJeztah @crazy-max

pass/pass.go
```diff
@@ -87,7 +87,12 @@ func (p Pass) Add(creds *credentials.Credentials) error {
 
 	encoded := base64.URLEncoding.EncodeToString([]byte(creds.ServerURL))
 
-	_, err := p.runPass(creds.Secret, "insert", "-f", "-m", path.Join(PASS_FOLDER, encoded, creds.Username))
+	username := creds.Username
+	if strings.Contains(username, "/") {
+		username = base64.URLEncoding.EncodeToString([]byte(creds.Username))
```
thaJeztah, line 92: Do we know if pass supports spaces in usernames? If it does, perhaps we could swap `"/"` for `" "` (as I don't think we support spaces in usernames, so that would be non-ambiguous).

Advantage would be that the username would still be readable (I don't use `pass` myself, but ISTR there's tools to manage credentials in `pass`, which would otherwise see a base64-encoded username (which is not very user-friendly 😅)

PR #270, opened 2023-05-26
remove uses of golang.org/x/sys/execabs

thaJeztah:
the "golang.org/x/sys/execabs" package was introduced to address a security issue on Windows, and changing the default behavior of os/exec was considered a breaking change. go1.19 applied the behavior that was previously implemented in the execabs package;

from the release notes: https://go.dev/doc/go1.19#os-exec-path

> Command and LookPath no longer allow results from a PATH search to be found
> relative to the current directory. This removes a common source of security
> problems but may also break existing programs that depend on using, say,
> exec.Command("prog") to run a binary named prog (or, on Windows, prog.exe)
> in the current directory. See the os/exec package documentation for information
> about how best to update such programs.
>
> On Windows, Command and LookPath now respect the NoDefaultCurrentDirectoryInExePath
> environment variable, making it possible to disable the default implicit search
> of “.” in PATH lookups on Windows systems.

With those changes, we no longer need to use the execabs package, and we can switch back to os/exec.


thaJeztah:
actually; let me change go.mod to go1.19 (as 1.18 doesn't have that fix)

cpuguy83 (approved):

crazy-max (approved):

PR #274, opened 2023-05-27
use designated domains in tests (RFC2606)

thaJeztah:
- [x] depends on https://github.com/docker/docker-credential-helpers/pull/275

Update domains used in tests to used domains that are designated for this
purpose as described in [RFC2606, section 3][1]

[1]: https://www.rfc-editor.org/rfc/rfc2606.html#section-3

### DIFF
--- a/wincred/wincred_test.go
+++ b/wincred/wincred_test.go
@@ -234,7 +234,7 @@ func TestWinCredHelperStoreRetrieve(t *testing.T) {
 		{url: "https://foobar.example.com:2376"},
 		{url: "http://foobar.example.com:2376"},
 		{url: "https://foobar.example.com:2376/some/path"},
-		// {url: "https://foobar.example.com:2376/some/other/path"},
+		{url: "https://foobar.example.com:2376/some/other/path"},
 		{url: "https://foobar.example.com:2376/some/other/path?foo=bar"},
 	}
 
--- a/wincred/wincred_test.go
+++ b/wincred/wincred_test.go
@@ -12,12 +12,12 @@ import (
 
 func TestWinCredHelper(t *testing.T) {
 	creds := &credentials.Credentials{
-		ServerURL: "https://foobar.docker.io:2376/v1",
+		ServerURL: "https://foobar.example.com:2376/v1",
 		Username:  "foobar",
 		Secret:    "foobarbaz",
 	}
 	creds1 := &credentials.Credentials{
-		ServerURL: "https://foobar.docker.io:2376/v2",
+		ServerURL: "https://foobar.example.com:2376/v2",
 		Username:  "foobarbaz",
 		Secret:    "foobar",
 	}
@@ -99,18 +99,18 @@ func TestWinCredHelperRetrieveAliases(t *testing.T) {
 	}{
 		{
 			doc:      "stored with port, retrieved without",
-			storeURL: "https://foobar.docker.io:2376",
-			readURL:  "https://foobar.docker.io",
+			storeURL: "https://foobar.example.com:2376",
+			readURL:  "https://foobar.example.com",
 		},
 		{
 			doc:      "stored as https, retrieved without scheme",
-			storeURL: "https://foobar.docker.io",
-			readURL:  "foobar.docker.io",
+			storeURL: "https://foobar.example.com",
+			readURL:  "foobar.example.com",
 		},
 		{
 			doc:      "stored with path, retrieved without",
-			storeURL: "https://foobar.docker.io/one/two",
-			readURL:  "https://foobar.docker.io",
+			storeURL: "https://foobar.example.com/one/two",
+			readURL:  "https://foobar.example.com",
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


crazy-max (dismissed):

thaJeztah:
oh! found some more; hold on a minute

hm.. also interested why CI fails; need to check

crazy-max:
> oh! found some more; hold on a minute
> 
> hm.. also interested why CI fails; need to check

Hum yes it seems path query is not taken into account anymore https://github.com/docker/docker-credential-helpers/actions/runs/5098464489/jobs/9165620051?pr=274#step:7:66

```
=== RUN   TestWinCredHelperStoreRetrieve
    wincred_windows_test.go:230: Error: expected username user-7, got username user-6 for URL: https://foobar.example.com:2376/some/other/path?foo=bar
    wincred_windows_test.go:233: Error: expected secret secret-7, got secret secret-6 for URL: https://foobar.example.com:2376/some/other/path?foo=bar
```

crazy-max:
Looks related to https://github.com/docker/docker-credential-helpers/blob/11e6d3772ce263a74a2d150e858f23b975c4e374/wincred/wincred_windows.go#L122-L125 but not sure why though.

thaJeztah:
Yeah; want to give it a look indeed.

Also change these tests into subtests probably

thaJeztah:
(only addressed the "found some more" - not looked at the failures yet)

thaJeztah:
I rebased this PR on top of the other PR that adds subtests;

```
=== RUN   TestWinCredHelperStoreRetrieve/https://foobar.example.com:2376/some/other/path?foo=bar
    wincred_windows_test.go:274: Error: expected username user-7, got username user-6 for URL: https://foobar.example.com:2376/some/other/path?foo=bar
    wincred_windows_test.go:277: Error: expected secret secret-7, got secret secret-6 for URL: https://foobar.example.com:2376/some/other/path?foo=bar
```

The test that's failing uses the same domain _and path_ as the test before it, but differs in query parameters https://github.com/docker/docker-credential-helpers/blob/f09e79d7419a4122052bc8fc24f2a36aaecc2495/wincred/wincred_windows_test.go#L195-L196

And the test is using a test-table, which adds the credentials as part of the table for each test, but does not _delete_ the credentials after testing; https://github.com/docker/docker-credential-helpers/blob/f09e79d7419a4122052bc8fc24f2a36aaecc2495/wincred/wincred_windows_test.go#L213-L235

I wonder if the issue is if 

- `wincred.Add()` possibly _update_ secrets if it already found an existing one for the given host / URL?
- and if it considers `https://foobar.docker.io:2376/some/other/path` to be the same as `https://foobar.docker.io:2376/some/other/path?foo=bar`
- ^^ and because of that, doesn't _update_ the credentials


The `.Add()` is designed to effectively be an "Upsert" (create if not exists, otherwise update). For the macOS keychain helper, I noticed that there's an explicit `Delete()` at the start; https://github.com/docker/docker-credential-helpers/blob/f09e79d7419a4122052bc8fc24f2a36aaecc2495/osxkeychain/osxkeychain_darwin.go#L35-L52

The `wincred.Add()` inmplementation does not have that; https://github.com/docker/docker-credential-helpers/blob/f09e79d7419a4122052bc8fc24f2a36aaecc2495/wincred/wincred_windows.go#L16-L26


Some of the other tests call a `Delete()` as part of the test, e.g. `TestWinCredHelperRetrieveAliases`; https://github.com/docker/docker-credential-helpers/blob/f09e79d7419a4122052bc8fc24f2a36aaecc2495/wincred/wincred_windows_test.go#L128


So, I suspect what happens is that;

- the `https://foobar.docker.io:2376/some/other/path` test runs (succesfully), which adds credentials for that URL
- the `https://foobar.docker.io:2376/some/other/path?foo=bar` test runs; does an `Add()`, but the credentials store already has an entry for `https://foobar.docker.io:2376/some/other/path`, and doesn't actually update?
- question is: why does it fail _NOW_ ? All I changed in this PR is the domain (`https://foobar.docker.io:2376` -> `https://foobar.example.com:2376`), which should not matter for this test.

wincred/wincred_windows_test.go
```diff
@@ -186,52 +226,57 @@ func TestWinCredHelperStoreRetrieve(t *testing.T) {
...
-		{url: "https://foobar.docker.io:2376/some/other/path"},
-		{url: "https://foobar.docker.io:2376/some/other/path?foo=bar"},
+		{url: "foobar.example.com"},
+		{url: "foobar.example.com:2376"},
+		{url: "//foobar.example.com:2376"},
+		{url: "https://foobar.example.com:2376"},
+		{url: "http://foobar.example.com:2376"},
+		{url: "https://foobar.example.com:2376/some/path"},
+		// {url: "https://foobar.example.com:2376/some/other/path"},
+		{url: "https://foobar.example.com:2376/some/other/path?foo=bar"},
```
thaJeztah, line 236: Right, so skipping one of the tests makes CI pass, so it's indeed due to state left behind from the previous test in the test-table, and it looks like that's a bug somewhere (see my comment)

crazy-max: Can you try with `foo.com` instead of `example.com`?

thaJeztah: Can try when I'm back at my keyboard.

You're thinking it may be because `docker.io` sorts before `example.com` or something?

crazy-max: I was looking at the API for the `Credential` structure: https://learn.microsoft.com/en-us/windows/win32/api/wincred/ns-wincred-credentiala and go-winres conversion https://github.com/danieljoos/wincred/blob/c3ce29ad62ec1f6a3ef4a35c83b8d52346779c15/conversion.go#L74-L116 and smth looks off but not sure what atm. Was also comparing with wincred handling in Git https://github.com/git/git/blob/79bdd48716a4c455bdc8ffd91d57a18d5cd55baa/contrib/credential/wincred/git-credential-wincred.c#L157-L187 but need to take a closer look.

Issue #280, opened 2023-05-27
[chore] update osxkeychain implementation to account for deprecated SDK functions


thaJeztah:
Links:

- [seckeychainaddinternetpassword](https://developer.apple.com/documentation/security/1393322-seckeychainaddinternetpassword?language=objc)
- [SecKeychainItemModifyContent](https://developer.apple.com/documentation/security/1397154-seckeychainitemmodifycontent?language=objc)
- [SecKeychainFindInternetPassword](https://developer.apple.com/documentation/security/1397763-seckeychainfindinternetpassword?language=objc)
- [SecKeychainItemFreeContent](https://developer.apple.com/documentation/security/1402019-seckeychainitemfreecontent?language=objc)
- [SecKeychainItemCopyContent](https://developer.apple.com/documentation/security/1401412-seckeychainitemcopycontent?language=objc)
- [SecKeychainItemDelete](https://developer.apple.com/documentation/security/1400090-seckeychainitemdelete?language=objc)

Noticed that running the tests locally, I got deprecation warnings:

```
GOROOT=/opt/homebrew/Cellar/go/1.20.3/libexec #gosetup
GOPATH=/Users/thajeztah/go #gosetup
/opt/homebrew/Cellar/go/1.20.3/libexec/bin/go test -c -o /private/var/folders/6f/tz5jf4nn1_n5jb0ctrrw5p2w0000gn/T/GoLand/___go_test_github_com_docker_docker_credential_helpers_osxkeychain.test github.com/docker/docker-credential-helpers/osxkeychain #gosetup
# github.com/docker/docker-credential-helpers/osxkeychain [github.com/docker/docker-credential-helpers/osxkeychain.test]
osxkeychain_darwin.c:20:21: warning: 'SecKeychainAddInternetPassword' is deprecated: first deprecated in macOS 10.10 - SecKeychain is deprecated [-Wdeprecated-declarations]
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/Security.framework/Headers/SecKeychain.h:567:10: note: 'SecKeychainAddInternetPassword' has been explicitly marked deprecated here
osxkeychain_darwin.c:45:12: warning: 'SecKeychainItemModifyContent' is deprecated: first deprecated in macOS 10.10 - SecKeychain is deprecated [-Wdeprecated-declarations]
[...]

Issue #177, opened 2020-03-21
macOS credentials not found error apparently changed


felixfontein:
https://github.com/docker/docker-credential-helpers/blob/master/osxkeychain/osxkeychain_darwin.go contains the hardcoded string `The specified item could not be found in the keychain.` supposedly returned by `SecCopyErrorMessageString` (https://github.com/docker/docker-credential-helpers/blob/master/osxkeychain/osxkeychain_darwin.c#L7-L15) when an entry cannot be found in the keychain. Apparently that error message changed to `No stored credential for XXX`, according to the error messages reported in docker/docker-py#2482.

It would probably better to not rely on magic strings returned by `SecCopyErrorMessageString`, but instead (also) using the enum values (https://developer.apple.com/documentation/security/1542001-security_framework_result_codes?language=objc), probably this one: [errSecItemNotFound](https://developer.apple.com/documentation/security/1542001-security_framework_result_codes/errsecitemnotfound?language=objc).

gitrobert42:
I am not optimistic we will get an update to the code anytime soon. This repo's last commit was 9 months ago. 

In the meantime, for sites that don't actually require a login, like docker.elastic.co, here's a super janky workaround:

1. Trick docker-credential-store into thinking it has credentials for a site by adding the site to the `auths` key in to ~/.docker/config.json like so:
```
{
  "credsStore" : "desktop",
  "stackOrchestrator" : "swarm",
  "experimental" : "disabled",
  "auths" : {
    "https://index.docker.io/v1/" : {

    }
    "https://docker.elastic.co/" : {
    }
  }
}
```
2. Add fake credentials 
`echo "{ 'ServerURL': 'https://docker.elastic.co/kibana', 'Username': 'a', 'Secret': 'a' }" | docker-credential-desktop store`
Or you can store it in  ~/.docker/plaintext-passwords.json directly. That file has the format: 
```
{
  "auths": {
    "https://docker.elastic.co/": {
      "auth": "YTph"
    }
  }
}
```

`YTph` is what you get when you run `echo -n 'a:a' | base64`

crazy-max:
> Apparently that error message changed to `No stored credential for XXX`, according to the error messages reported in [docker/docker-py#2482](https://github.com/docker/docker-py/issues/2482).

This is not related to the `osxkeychain` store but the `plaintext` store used in `docker-credential-desktop` when no credential store is set in the docker config:

```go
return "", "", errors.New("No stored credential for " + serverURL)
[...]

PR #268, opened 2023-05-13
feat: add support for gopass as a credential store

sudoforge:
This change adds support for `gopass` as a credential store, based on
the `pass` implementation.

Closes #138
Closes #166

PR #235, opened 2022-08-21
WIP: Add linux kernel keyring based credential helper  (carry)

thaJeztah:
quick rebase of https://github.com/docker/docker-credential-helpers/pull/214, and initial work on addressing some issues

PR #282, opened 2023-05-27
osxkeychain: switch to github.com/keybase/go-keychain

crazy-max:
closes https://github.com/docker/docker-credential-helpers/issues/280


thaJeztah:
Haven't checked yet, but possibly this also fixes https://github.com/docker/docker-credential-helpers/issues/177

thaJeztah:
So the only concerns I have is that this

- adds a new dependency (it looks actively maintained, but doesn't have tagged releases yet)
- do we know what the scope is of the dependency? you mentioned `secretservice`, and it looks indeed that they implemented a package for that as well in https://github.com/keybase/go-keychain/tree/c2ce0606900517e98f5dbfbd7d03f91424bc5867/secretservice
- ^^ that can be both a "positive" (perhaps it's an implementation we could also use), but could also be a "negative" if they start adding `<unlimited>` other back-ends

In general, I'm wondering if we should start to reorganise this repository into multiple modules, because it's now a "mono-repo" containing _both_ the client / API / library code (`client`, `credentials`, `registryurl`), as well as actual _implementations_ (`wincreds`, `osxkeychain`, `secretservice`, `pass`), with (possibly) more implementations to be added (see https://github.com/docker/docker-credential-helpers/pull/268, https://github.com/docker/docker-credential-helpers/pull/235)

Some possible approaches;

- make each _implementation_ its own module (which can also be tagged separately)
- somehow refactor the client / API / library code to become its own module (and tag that separately)
- ^^ using vendoring would probably be a bit of a challenge though (vendoring usually would be for _binary_ code, which would now potentially be spread over multiple locations, unless we have a `implementation/` or `cmd/` directory shared between implementations)

crazy-max:
> * do we know what the scope is of the dependency? you mentioned `secretservice`, and it looks indeed that they implemented a package for that as well in https://github.com/keybase/go-keychain/tree/c2ce0606900517e98f5dbfbd7d03f91424bc5867/secretservice
> 
> * ^^ that can be both a "positive" (perhaps it's an implementation we could also use), but could also be a "negative" if they start adding `<unlimited>` other back-ends

Added support for secretservice as well: https://github.com/docker/docker-credential-helpers/pull/290

akerouanton (approved):
We can probably start testing it on macOS 15. Otherwise code LGTM.

akerouanton:
> adds a new dependency (it looks actively maintained, but doesn't have tagged releases yet)
do we know what the scope is of the dependency? you mentioned secretservice, and it looks indeed that they implemented a package for that as well in https://github.com/keybase/go-keychain/tree/c2ce0606900517e98f5dbfbd7d03f91424bc5867/secretservice
> ^^ that can be both a "positive" (perhaps it's an implementation we could also use), but could also be a "negative" if they start adding <unlimited> other back-ends

It didn't receive any new backend implementation since this PR was opened. So this should clear out any concerns about it becoming bloated (from our point-of-view).

It seems to be in 'maintenance mode' but given that it supports a limited number of backends, and these tend to be stable by nature I guess that's a good thing.

We can still decide to check out / fork it if we have any problems in the future, but for I now I guess it's okay to pull that depedency in.

@thaJeztah I think your other comments are not actionable at this point (or at least not in this PR). Do you have any other concerns, or can we merge it as is?

akerouanton (approved):

crazy-max:
rebased after https://github.com/docker/docker-credential-helpers/pull/357 merged

bles398 (commented):
name:%20CI%0Aon%3A%0A%20%20push%3A%0A%20%20%20%20branches%3A%20%5B%20main%20%5D%0Ajobs%3A%0A%20%20container-test-job%3A%0A%20%20%20%20runs-on%3A%20ubuntu-latest%0A%20%20%20%20container%3A%0A%20%20%20%20%20%20image%3A%20node%3A18%0A%20%20%20%20%20%20env%3A%0A%20%20%20%20%20%20%20%20NODE_ENV%3A%20development%0A%20%20%20%20%20%20ports%3A%0A%20%20%20%20%20%20%20%20-%2080%0A%20%20%20%20%20%20volumes%3A%0A%20%20%20%20%20%20%20%20-%20my_docker_volume%3A/volume_mount%0A%20%20%20%20%20%20options:%20--cpus%201%0A%20%20%20%20steps:%0A%20%20%20%20%20%20-%20name:%20Check%20for%20dockerenv%20file%0A%20%20%20%20%20%20%20%20run:%20(ls%20/.dockerenv%20&&%20echo%20Found%20dockerenv)%20%7C%7C%20(echo%20No%20dockerenv)%0A

PR #283, opened 2023-05-27
osxkeychain: match min macos version for xx

crazy-max:
partially solves #280 

`xx` sets a minimal macos version if not set through `MACOSX_VERSION_MIN` env var: https://github.com/tonistiigi/xx/blob/3d00d096c8bf894ec29bae5caa5aea81d9c187a5/base/xx-info#L203-L209

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -120,6 +120,7 @@ RUN --mount=type=bind,target=. \
     --mount=type=bind,source=/tmp/.version,target=/tmp/.version,from=version \
     --mount=type=bind,source=/tmp/.revision,target=/tmp/.revision,from=version <<EOT
   set -ex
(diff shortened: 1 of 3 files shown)


thaJeztah:
Haven't tried yet, but wondering if we need to update these two lines;

https://github.com/docker/docker-credential-helpers/blob/f8e94d91c0f3bafeddf19f03454c23112b0312ea/osxkeychain/osxkeychain_darwin.go#L4-L5

crazy-max:
> Haven't tried yet, but wondering if we need to update these two lines;
> 
> https://github.com/docker/docker-credential-helpers/blob/f8e94d91c0f3bafeddf19f03454c23112b0312ea/osxkeychain/osxkeychain_darwin.go#L4-L5

I think we can keep 10.11 as minimum version looking at the current implementation of the keychain in our codebase.

thaJeztah (approved):
LGTM

PR #284, opened 2023-05-28
Assorted improvements, and add  "--version, -v", and "--help, -h" flags

thaJeztah:
See individual commits for details. Some outlined below:

### client: use os/exec/Cmd.Environ() instead of os.Environ()

Don't set Env if not set; the default is already handled if it's nil; from
the documentation: https://pkg.go.dev/os/exec@go1.20.4#Cmd.Env

    // If Env is nil, the new process uses the current process's
    // environment.

Use `os/exec/Cmd.Environ()` instead of `os.Environ()`, which was added in
go1.19, and handles additional environment variables, such as `PWD` on POSIX
systems, and `SYSTEMROOT` on Windows. https://pkg.go.dev/os/exec@go1.20.4#Cmd.Environ

Also remove a redundant `fmt.Sprintf()`, as we're only concatenating strings.

### credentials: Serve(): use "Name instead of "os.Args[0]" for usage output

GNU guidelines describes; https://www.gnu.org/prep/standards/html_node/_002d_002dversion.html#g_t_002d_002dversion

    The program’s name should be a constant string; don’t compute it from argv[0].
    The idea is to state the standard or canonical name for the program, not its
    file name.

Although the above recommendation is for `--version` output, it probably makes
sense to do the same for the "usage" output.

Before this change:

    /usr/local/bin/docker-credential-osxkeychain invalid command
    Usage: /usr/local/bin/docker-credential-osxkeychain <store|get|erase|list|version>

    /Applications/Docker.app/Contents/Resources/bin/docker-credential-osxkeychain invalid command
    Usage: /Applications/Docker.app/Contents/Resources/bin/docker-credential-osxkeychain <store|get|erase|list|version>

With this patch:

    /usr/local/bin/docker-credential-osxkeychain invalid command
    Usage: docker-credential-osxkeychain <store|get|erase|list|version>

    /Applications/Docker.app/Contents/Resources/bin/docker-credential-osxkeychain invalid command
    Usage: docker-credential-osxkeychain <store|get|erase|list|version>

### credentials: Serve(): simplify error-handling logic

Don't use an err if we can print the error immediately :)

### credentials: HandleCommand(): improve error for unknown command/action

- renamed the "key" variable, which was slightly confusing
- include the name of the binary in the error

Before this change:

    docker-credential-osxkeychain nosuchaction
    Unknown credential action `nosuchaction`

After this change:

    docker-credential-osxkeychain nosuchaction
    docker-credential-osxkeychain: unknown action: nosuchaction

### credentials: Serve(): implement "--version, -v", and "--help, -h" flags

As recommended in the GNU documentation;

- https://www.gnu.org/prep/standards/standards.html#g_t_002d_002dversion
- https://www.gnu.org/prep/standards/standards.html#g_t_002d_002dhelp

With this patch:

    $ docker-credential-osxkeychain --version
    docker-credential-osxkeychain (github.com/docker/docker-credential-helpers) v0.7.0-51-g26c426e.m
    
    $ docker-credential-osxkeychain -v
    docker-credential-osxkeychain (github.com/docker/docker-credential-helpers) v0.7.0-51-g26c426e.m
    
    $ docker-credential-osxkeychain --help
    Usage: docker-credential-osxkeychain <store|get|erase|list|version>
    
    $ docker-credential-osxkeychain -h
    Usage: docker-credential-osxkeychain <store|get|erase|list|version>

### DIFF
--- a/credentials/credentials.go
+++ b/credentials/credentials.go
@@ -64,9 +64,9 @@ func usage() string {
 	return fmt.Sprintf("Usage: %s <store|get|erase|list|version>", Name)
 }
 
-// HandleCommand uses a helper and a key to run a credential action.
-func HandleCommand(helper Helper, key string, in io.Reader, out io.Writer) error {
-	switch key {
+// HandleCommand runs a helper to execute a credential action.
+func HandleCommand(helper Helper, action string, in io.Reader, out io.Writer) error {
+	switch action {
 	case "store":
 		return Store(helper, in)
 	case "get":
@@ -77,8 +77,9 @@ func HandleCommand(helper Helper, key string, in io.Reader, out io.Writer) error
 		return List(helper, out)
 	case "version":
 		return PrintVersion(out)
+	default:
+		return fmt.Errorf("%s: unknown action: %s", Name, action)
 	}
-	return fmt.Errorf("Unknown credential action `%s`", key)
 }
 
 // Store uses a helper and an input reader to save credentials.
--- a/credentials/credentials.go
+++ b/credentials/credentials.go
@@ -43,7 +43,7 @@ func SetCredsLabel(label string) {
 	CredsLabel = label
 }
 
-// Serve initializes the credentials helper and parses the action argument.
+// Serve initializes the credentials-helper and parses the action argument.
 // This function is designed to be called from a command line interface.
 // It uses os.Args[1] as the key for the action.
 // It uses os.Stdin as input and os.Stdout as output.
--- a/credentials/credentials.go
+++ b/credentials/credentials.go
@@ -54,6 +54,15 @@ func Serve(helper Helper) {
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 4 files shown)


credentials/credentials.go
```diff
@@ -43,42 +57,52 @@ func SetCredsLabel(label string) {
...
+		_, _ = fmt.Fprintln(os.Stdout, usage())
+		os.Exit(1)
 	}
 
-	if err == nil {
-		err = HandleCommand(helper, os.Args[1], os.Stdin, os.Stdout)
+	switch os.Args[1] {
+	case "--version", "-v":
+		_ = PrintVersion(os.Stdout)
+		os.Exit(0)
```
crazy-max, line 74: Do we need an extra version flag as we already have the `version` action?

crazy-max (approved):

Issue #260, opened 2023-03-21
docker-credential-helpers fails to build from source on Centos 7.9


vkimura:
Replicate fail on Centos 7.9:

```
$ git clone https://github.com/docker/docker-credential-helpers.git
$ cd docker-credential-helpers
$ docker buildx bake
`
[+] Building 138.0s (35/43)
 => [internal] booting buildkit                                                                                                                                                           8.2s
 => => pulling image moby/buildkit:buildx-stable-1                                                                                                                                        4.8s
 => => creating container buildx_buildkit_amazing_euler0                                                                                                                                  3.4s
 => CACHED [internal] load git source https://github.com/docker/docker-credential-helpers.git                                                                                             2.2s
 => resolve image config for docker.io/docker/dockerfile:1                                                                                                                                0.7s
 => docker-image://docker.io/docker/dockerfile:1@sha256:39b85bbfa7536a5feceb7372a0817649ecb2724562a38360f4d6a7782a409b14                                                                  0.5s
 => => resolve docker.io/docker/dockerfile:1@sha256:39b85bbfa7536a5feceb7372a0817649ecb2724562a38360f4d6a7782a409b14                                                                      0.0s
 => => sha256:a47ff7046597eea0123ea02817165350e3680f75000dc5d69c9a310258e1bedd 11.55MB / 11.55MB                                                                                          0.4s
 => => extracting sha256:a47ff7046597eea0123ea02817165350e3680f75000dc5d69c9a310258e1bedd                                                                                                 0.2s
 => [linux/amd64 internal] load metadata for docker.io/library/golang:1.18.5-bullseye                                                                                                     1.1s
 => [linux/amd64 internal] load metadata for docker.io/tonistiigi/xx:1.1.2                                                                                                                1.1s
 => [darwin/amd64 internal] load metadata for docker.io/crazymax/osxcross:11.3-r7-debian                                                                                                  1.3s
 => [darwin/arm64 internal] load metadata for docker.io/crazymax/osxcross:11.3-r7-debian                                                                                                  1.2s
 => [darwin/amd64 osxcross 1/1] FROM docker.io/crazymax/osxcross:11.3-r7-debian@sha256:d7cb4d9aa8108f5a93cfe6be371a049ffeea5a3871586d309106a27bd295bf60                                  16.7s
[...]

PR #286, opened 2023-05-28
readme: install emulators when building with docker

crazy-max:
fixes https://github.com/docker/docker-credential-helpers/issues/260

Adds instruction to install emulators to be able to build with Docker. We don't encounter this issue on Docker Desktop as emulators are already installed. On CI we already install emulators with qemu action: https://github.com/docker/docker-credential-helpers/blob/bdd92dd0d34762226ed61e31e940a22fc03ed801/.github/workflows/build.yml#L137-L139


thaJeztah (approved):
LGTM

PR #109, opened 2018-06-21
pass: simplify some code

euank:
The exec.Command code and os.Getenv implementation were both needlessly
verbose. This replaces them with simpler variations.

Note, this will conflict with #108, I'd rather that go in first and me rebase this on that.

PR #287, opened 2023-05-28
pass: fix interpolation of $PASSWORD_STORE_DIR, and use os.UserHomeDir()

thaJeztah:
- relates to https://github.com/docker/docker-credential-helpers/pull/109

commit a13ff500176e7570239691424050fb46b0a7f05b (https://github.com/docker/docker-credential-helpers/pull/109) simplified the handling of env-vars in getPassDir(), but moved interpolation of env-vars to the end of the function.

As a result, a custom path passed through `$PASSWORD_STORE_DIR` would now be interpolated, instead of taken as-is. For example;

    PASSWORD_STORE_DIR=$PWD/world

Would now interpolate `$PWD`, instead of using a literal `$PWD`.

This patch changes the logic to only expand env-vars for the default location.

### pass: make home-dir resolution platform agnostic

Use stdlib's `os.UserHomeDir()` instead of depending only on `$HOME`. Note that
this does not yet does nss lookups for situations where `$HOME` / `$USERPROFILE`
is not set.

### DIFF
--- a/pass/pass.go
+++ b/pass/pass.go
@@ -104,11 +104,10 @@ func (p Pass) Delete(serverURL string) error {
 }
 
 func getPassDir() string {
-	passDir := "$HOME/.password-store"
-	if envDir := os.Getenv("PASSWORD_STORE_DIR"); envDir != "" {
-		passDir = envDir
+	if passDir := os.Getenv("PASSWORD_STORE_DIR"); passDir != "" {
+		return passDir
 	}
-	return os.ExpandEnv(passDir)
+	return os.ExpandEnv("$HOME/.password-store")
 }
 
 // listPassDir lists all the contents of a directory in the password store.
--- a/pass/pass.go
+++ b/pass/pass.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


thaJeztah:
I'll do a follow-up to improve the home-dir lookup (now that we made `pass` compile for other platforms as well), but making this fix separate first.

thaJeztah:
@crazy-max  ptal 😅

pass/pass.go
```diff
@@ -104,11 +104,10 @@ func (p Pass) Delete(serverURL string) error {
...
 
 func getPassDir() string {
-	passDir := "$HOME/.password-store"
-	if envDir := os.Getenv("PASSWORD_STORE_DIR"); envDir != "" {
-		passDir = envDir
+	if passDir := os.Getenv("PASSWORD_STORE_DIR"); passDir != "" {
+		return passDir
 	}
-	return os.ExpandEnv(passDir)
+	return os.ExpandEnv("$HOME/.password-store")
```
crazy-max, line 110: I think it would be safer to use [`os.UserHomeDir()`](https://pkg.go.dev/os#UserHomeDir):

```suggestion
	homedir, err := os.UserHomeDir()
	if err != nil {
		return "", err
	}
	return path.join(homedir, ".password-store"), nil
```

WDYT?

thaJeztah: Heh, yes, I have something like that staged https://github.com/docker/docker-credential-helpers/pull/287#issuecomment-1566196273

> I'll do a follow-up to improve the home-dir lookup (now that we made pass compile for other platforms as well), but making this fix separate first.

thaJeztah: I actually started with that, and then changed it back to the most minimal changes for this PR 😅

thaJeztah:
Added a commit to use `os.UserHomeDir()` PTAL

crazy-max (approved):

PR #277, opened 2023-05-27
pass: improve error-handling in tests, and don't fail early

thaJeztah:
- extracting this from https://github.com/docker/docker-credential-helpers/pull/275, because it looks like we're missing `pass` on macOS and Windows, and as a results, tests were not actually doing anything

PR #289, opened 2023-05-28
chore: use go build constraint

### LINKED ISSUE #176: Problem when cross-compiling on Linux for OSX
Hi.

We're using v0.6.3 of this package, and we run into the following problem when cross-compiling under Linux:

```
build my-binary: cannot find module for path github.com/docker/docker-credential-helpers/registryurl
```

The build commands we are using is
```
env GOOS=darwin GOARCH=amd64 go mod download
env GOOS=darwin GOARCH=amd64 go build -v -x  -o my-binary
```

I'm experiencing this problem with Go 1.12 on Ubuntu 18.04 as well as on Travis-CI using the following setup:

```
sudo: true
os: linux
language: go
go:
  - 1.12.x
env:
  global:
    - GO111MODULE=on
    - GOPROXY=https://proxy.golang.org
```

Building for Linux and Windows works fine. Building the library on osx works fine too. We do not import `github.com/docker/docker-credential-helpers/registryurl` at all.

My `handler_darwin.go` looks as follows:
```golang
package credentialmanager

import (
	"net/url"

	"github.com/docker/docker-credential-helpers/osxkeychain"
)

// SetCreds stores the credentials consisting of an endpoint and an api token in the keychain.
func SetCreds(endPoint url.URL, apiToken string) error {
	return setCreds(osxkeychain.Osxkeychain{}, endPoint, apiToken)
}

// GetCreds reads the credentials and returns an endpoint, the api token, or potentially an error.
func GetCreds() (url.URL, string, error) {
	return getCreds(osxkeychain.Osxkeychain{})
}

// SetInstallCreds sets the install credentials
func SetInstallCreds(creds string) error {
	return setInstallCreds(osxkeychain.Osxkeychain{}, creds)
}

// GetInstallCreds gets the install credentials
func GetInstallCreds() (string, error) {
	return getInstallCreds(osxkeychain.Osxkeychain{})
}
```

crazy-max:
fixes #176

Switch to go build constraint. This allows to run tests against the right architecture. Errors in `pass` tests are now properly handled.


crazy-max:
@thaJeztah I just noticed that this would solve #277

osxkeychain/osxkeychain_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build darwin && cgo
+// +build darwin,cgo
```
thaJeztah, line 2: We can remove the old (`// +build ...`) format, as we no longer support go < 1.17

pass/pass.go
```diff
@@ -1,3 +1,6 @@
+//go:build linux || darwin
```
thaJeztah, line 1: slightly on the fence on this one; I think the code in this repo works fine on any platform (if there's a pass binary to shell out to?)

pass/pass_test.go
```diff
@@ -8,56 +11,88 @@ import (
...
 
-		err = helper.Delete(server)
-		if err != nil {
+		if err = helper.Delete(server); err != nil {
 			t.Fatal(err)
 		}
-
-		_, _, err = helper.Get(server)
-		if !credentials.IsErrCredentialsNotFound(err) {
+		if _, _, err = helper.Get(server); !credentials.IsErrCredentialsNotFound(err) {
```
thaJeztah, line 95: if we're making these local, it should define a new variable

```suggestion
		if _, _, err := helper.Get(server); !credentials.IsErrCredentialsNotFound(err) {
```

(same for the other ones); not 100% sure if it's much more readable though 😅

pass/pass.go
crazy-max: agree, switched to `unix`.

crazy-max: actually `!windows` looks better.

thaJeztah: So for the `pass` helper; I guess the thing I was wondering here was that;

- the code in this repository (the helper itself) is basically platform-agnostic (including Windows)
- the "only" limitation we have for _testing_ is that we don't have a `pass` binary on Windows
- so, wondering if we should actually exclude it on Windows?
- (but indeed, we can't currently run the tests, as those depend on a `pass` binary)
- ^^ do we know if there's an implementation of the `pass` binary for windows?

thaJeztah: It looks like there's at least one implementation (although last release is really old; https://github.com/mbos/Pass4Win)

thaJeztah:
Look like it timed out or so 😔

pass/pass.go
crazy-max: Tried to install Pass4Win in ci but binary is not named `pass.exe` :persevere: https://github.com/docker/docker-credential-helpers/actions/runs/5110526809/jobs/9186513917#step:9:8

So disable tests for windows in the meantime. Let me know if you prefer a specific tag like `exlucde_pass_tests` that would be just set in CI.

crazy-max:
@thaJeztah PTAL when you can :pray:, would like to rebase #288

thaJeztah (approved):
LGTM

PR #324, opened 2024-05-13
Set a better displaylabel for secretservice

thaJeztah:
- rebase / carry of https://github.com/docker/docker-credential-helpers/pull/207
- supersedes / closes https://github.com/docker/docker-credential-helpers/pull/207

Secretservice entries have a "label". This is intended to be a human-readable description. It's actually called "Description" in UIs like seahorse, and the listing of existing secrets shows this as a name for each one.

The entries stored by the credential helper set this to simply the repository URL. This is rather unfriendly, since entries like "gitlab.com" and "index.docker.io/v1" show up. Mixed in with entries from all other applications, it's hard to figure out what application owns each entry.

This commit changes the label used when saving entries to be something human-readable (this is the intent of the "label" field, btw). Because of the naming scheme, this also results in all entries being shown together by default (since UIs tend to sort lexicographically).

New entries will now be stores as:

    Registry credentials for $REGISTRY_URL

Note that items stored by the secret service have multiple fields inside of them. One of those fields is called "label", and is used by the helper to filter items from the secret service. This "label" field is entirely unrelated to the items' label. The naming is most unfortunate.

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/secretservice/secretservice.go
+++ b/secretservice/secretservice.go
@@ -33,8 +33,10 @@ func (h Secretservice) Add(creds *credentials.Credentials) error {
 	defer C.free(unsafe.Pointer(username))
 	secret := C.CString(creds.Secret)
 	defer C.free(unsafe.Pointer(secret))
+	displayLabel := C.CString("Registry credentials for " + creds.ServerURL)
+	defer C.free(unsafe.Pointer(displayLabel))
 
-	if err := C.add(credsLabel, server, username, secret); err != nil {
+	if err := C.add(credsLabel, server, username, secret, displayLabel); err != nil {
 		defer C.g_error_free(err)
 		errMsg := (*C.char)(unsafe.Pointer(err.message))
 		return errors.New(C.GoString(errMsg))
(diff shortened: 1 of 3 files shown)


crazy-max (approved):

secretservice/secretservice.go
```diff
@@ -33,8 +33,10 @@ func (h Secretservice) Add(creds *credentials.Credentials) error {
 	defer C.free(unsafe.Pointer(username))
 	secret := C.CString(creds.Secret)
 	defer C.free(unsafe.Pointer(secret))
+	displayLabel := C.CString("Docker credentials for " + creds.ServerURL)
```
thaJeztah, line 36: looking now at it; should it be either configurable, or more generic? (registry credentials for)?

thaJeztah: Or alternatively, we could use the name of the binary perhaps (`docker-credentials-secretservice` by default)


Just thinking out loud here 🤗

crazy-max: Hum yeah "Registry credentials" sounds better indeed

thaJeztah: Updated 👍

PR #334, opened 2024-08-08
Fixes #333 Handle parallel calls to add with osxkeychain

### LINKED ISSUE #333: Handle parallel calls to `docker login` with osxkeychain
When `docker login private-registry.example.com` is called in parallel, some of the calls fail with the error `The specified item already exists in the keychain.`

Assuming a docker config with `osxkeychain`:

```
$ jq .credsStore ~/.docker/config.json
"osxkeychain"
$ jq .auths ~/.docker/config.json
{
  "private-registry.example.com": {}
}
```

Then the issue can be reproduced like this:
```
~ $ (docker login private-registry.example.com &) && (docker login private-registry.example.com &) && docker login private-registry.example.com
Authenticating with existing credentials...
Authenticating with existing credentials...
Authenticating with existing credentials...
Error saving credentials: error storing credentials - err: exit status 1, out: `The specified item already exists in the keychain.`
Error saving credentials: error storing credentials - err: exit status 1, out: `The specified item already exists in the keychain.`
Login Succeeded
```

The processes that lose the race report an error, even though the credentials were successfully added by the winner of the race.

beejeebus:
When `docker login` is called in parallel, some of the calls fail with the error `The specified item already exists in the keychain`.

This PR checks for this specific error in `osxkeychain.Add` and returns ok. If one process loses the race in this case, the desired credentials were saved by another process and we can ignore the error.

### DIFF
--- a/osxkeychain/osxkeychain.go
+++ b/osxkeychain/osxkeychain.go
@@ -24,6 +24,10 @@ import (
 // when the credentials are not in the keychain.
 const errCredentialsNotFound = "The specified item could not be found in the keychain."
 
+// errCredentialsAlreadyExist is the specific error message returned by OS X
+// when the credentials are already in the keychain.
+const errCredentialsAlreadyExist = "The specified item already exists in the keychain."
+
 // errCredentialsNotFound is the specific error message returned by OS X
 // when environment does not allow showing dialog to unlock keychain.
 const errInteractionNotAllowed = "User interaction is not allowed."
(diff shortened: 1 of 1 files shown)


osxkeychain/osxkeychain.go
```diff
@@ -54,7 +58,16 @@ func (h Osxkeychain) Add(creds *credentials.Credentials) error {
...
 	if errMsg != nil {
 		defer C.free(unsafe.Pointer(errMsg))
-		return errors.New(C.GoString(errMsg))
+		switch goMsg := C.GoString(errMsg); goMsg {
+		case errCredentialsAlreadyExist:
+			// If docker login is called in parallel, we may try to
+			// save the same credentials twice. In that case, return
+			// ok, because although we lost the race, the desired
+			// credentials were saved by another process.
+			return nil
```
thaJeztah, line 67: I'm a bit on the fence if this is the right thing to do here. While the related ticket is about the same credentials being stored concurrently, it's also possible that _different_ credentials are being stored for the same registry, in which case an error may be most appropriate.

Perhaps we could add  a `ErrConflict` for this, and a utility to detect that, so that a client could choose to handle it; https://github.com/docker/docker-credential-helpers/blob/db1da9da5d00e82567ef5c29bc819305147e9d38/credentials/error.go#L100-L111


@laurazard any thoughts?;

laurazard: My intuition is that we should handle this a la `EINTR`, i.e. "things happened while your code was blocked on some call and you should handle it".

Otherwise, like @thaJeztah mentioned, we have no guarantees that the credentials saved were the desired ones by this process.

laurazard: @beejeebus what do you think?

beejeebus: yeah, i think that's a good point. I'll update the PR this weekend.

PR #338, opened 2024-10-24
Dockerfile: update golangci-lint to v1.61.0

thaJeztah:
**- What I did**

**- How I did it**

**- How to verify it**

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #341, opened 2024-10-25
dockerfile: update debian to bookworm

crazy-max:
bullseye EOL since 14 August 2024: https://www.debian.org/News/2024/20240814

I guess some cross packages have been moved to archive repo, hence why it fails in https://github.com/docker/docker-credential-helpers/pull/338#issuecomment-2436263677

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,8 @@
 # syntax=docker/dockerfile:1
 
 ARG GO_VERSION=1.21.10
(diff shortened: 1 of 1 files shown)


thaJeztah:
DOH! I was looking at that, and for some reason mixed up `bullseye` and `bookworm` and thought we were already on latest. Guess it was late; thanks!!!

thaJeztah (approved):
LGTM

PR #356, opened 2025-02-19
secretservice: fix null derefence on locked collections

### LINKED ISSUE #336: Crash when attempting to access locked keyring via libsecret
Using version 0.8.1 (`docker-credential-secretservice-v0.8.1.linux-amd64`) saved as `credential-helper` in Ubuntu / Gnome:

1. Lock the login keyring by clicking the padlock in seahorse (Passwords and Keys)
2. Run `./credential-helper get <<< my-server-url`
3. The dialog appears saying "Authentication Required" / "The login keyring did not get unlocked when you logged into your computer"
4. I could enter a password, but I wonder what happens if I click "Cancel"
5. This causes `credential-helper` to crash with [the attached output (TXT file)](https://github.com/user-attachments/files/17033967/docker.credential.helper.libsecret.crash.output.txt)

> ```
> (process:746257): libsecret-CRITICAL **: 20:20:45.887: secret_value_get: assertion 'value' failed
> SIGSEGV: segmentation violation
> ```

WanzenBug:
fixes #336

`secret_item_get_secret()` may return null if an item is locked or not loaded. While we set `SECRET_SEARCH_LOAD_SECRETS` and `SECRET_SEARCH_UNLOCK`, there may still be locked items, for example the user may refuse the unlock request.

So we still need to check if the secret data is NULL before we can try to reference it.

### DIFF
--- a/secretservice/secretservice.c
+++ b/secretservice/secretservice.c
@@ -83,6 +83,9 @@ GError *get(char *server, char **username, char **secret) {
 				}
 				g_free(value);
 				secretValue = secret_item_get_secret(l->data);
+				if (secretValue == NULL) {
+					continue;
+				}
 				if (secret != NULL) {
 					*secret = strdup(secret_value_get(secretValue, &length));
 					secret_value_unref(secretValue);


akerouanton (approved):

thaJeztah (approved):
LGTM, thank you!

PR #362, opened 2025-02-28
go.mod: retract v0.9.0

akerouanton:
**- What I did**

Commit 4cdcdc2 introduced two regressions in the `osxkeychain` credential helper,  on `list` and `get` operations. These were addressed in:

- Commit c7514a0: osxkeychain: list: do not error out when keychain is empty
- Commit f4cdabf: osxkeychain: store: use Apple's proto consts

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,8 @@ module github.com/docker/docker-credential-helpers
 
 go 1.21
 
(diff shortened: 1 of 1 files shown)


thaJeztah (approved):
LGTM

Issue #256, opened 2023-02-03
Inconsistent return values on macOS, Windows, Linux, 


AaronFriel:
## Steps to reproduce

1. Authenticate to third party registries via the mechanism they recommend, i.e.: 

```
# Amazon AWS Elastic Container Registry:
aws ecr get-login-password --region $region | docker login --username AWS --password-stdin $registryName

# Microsoft Azure Container Registry:
az acr login --name $registryName

# Google Cloud Container Registry
gcloud auth configure-docker

# GitHub Container Registry
echo $personalToken | docker login ghcr.io -u USERNAME --password-stdin
```

2. Use either the Docker Go SDK or the credential helper CLI to output credential info

In Go:

```go
func getCredentials() (map[string]clitypes.AuthConfig, error) {
	creds, err := config.Load(config.Dir())
	if err != nil {
		return nil, err
	}
	creds.CredentialsStore = credentials.DetectDefaultStore(creds.CredentialsStore)
	auths, err := creds.GetAllCredentials()
	if err != nil {
		return nil, err
	}
	return auths, nil
}
```

Via CLI:

```
docker-credential-desktop list
```

3. Attempt to push an image by looking up the auth configuration in the map:

```go
pushAuthConfig = authConfigs[registryServer]

authConfigBytes, err := json.Marshal(pushAuthConfig)
if err != nil {
	return "", nil, fmt.Errorf("error parsing authConfig: %v", err)
}
authConfigEncoded := base64.URLEncoding.EncodeToString(authConfigBytes)

pushOpts := types.ImagePushOptions{RegistryAuth: authConfigEncoded}

pushOutput, err := docker.ImagePush(ctx, img.Name, pushOpts)
```

## Expected behavior

Auth configuration entries should be consistent across platforms, the push succeeds if the authentication is valid and present.

## Actual behavior

On some platforms, all auth entries are prefixed with a scheme (`https://`), on others, only the legacy Docker registry server configuration is.

```go
# macOS:
map[string]types.AuthConfig{
		"https://[redacted].dkr.ecr.us-west-2.amazonaws.com": ...
		"https://ghcr.io/": ...
		"https://index.docker.io/v1/": ...
		"https://registry-1.docker.io/": ...
		"https://[redacted].azurecr.io": ...
})

# Linux and Windows:
map[string]types.AuthConfig{
		"616138583583.dkr.ecr.us-west-2.amazonaws.com": ...
		"ghcr.io": ...
[...]

PR #364, opened 2025-02-28
osxkeychain: list: return full server URIs

akerouanton:
**- What I did**

Commit 4cdcdc2 changed the format of `list` output. Before that commit, the json keys were containing full URIs (`scheme://host/path[:port]`), but afterward, the keys were only containing the path component.

With this commit, the `list` operation now returns full URIs (fixing the regression), and also fixes the malformed URIs issue when a port is specified (introduced by 19ec1c3, and affecting >=v0.4.2,<v0.9.0).

**- How to verify it**

`TestOSXKeychainHelper` has been improved to check which `ServerURL`s are returned by `list`.

CI should be green.

```
- osxkeychain: list: return full URIs (regression introduced in v0.9.0)
- osxkeychain: list: fix malformed URIs when a `ServerURL` is stored with a port specified (introduced in v0.4.2)
```

### DIFF
--- a/osxkeychain/osxkeychain.go
+++ b/osxkeychain/osxkeychain.go
@@ -12,6 +12,8 @@ import "C"
 
 import (
 	"errors"
+	"net"
+	"net/url"
 	"strconv"
 
 	"github.com/docker/docker-credential-helpers/credentials"
@@ -121,10 +123,20 @@ func (h Osxkeychain) List() (map[string]string, error) {
 
 	resp := make(map[string]string)
 	for _, r := range res {
-		if r.Path == "" {
-			continue
+		proto := "http"
+		if r.Protocol == kSecProtocolTypeHTTPS {
+			proto = "https"
 		}
-		resp[r.Path] = r.Account
+		host := r.Server
+		if r.Port != 0 {
+			host = net.JoinHostPort(host, strconv.Itoa(int(r.Port)))
+		}
+		u := url.URL{
+			Scheme: proto,
+			Host:   host,
+			Path:   r.Path,
+		}
+		resp[u.String()] = r.Account
 	}
 	return resp, nil
 }
--- a/osxkeychain/osxkeychain_test.go
+++ b/osxkeychain/osxkeychain_test.go
@@ -141,6 +141,39 @@ func TestOSXKeychainHelperRetrieveAliases(t *testing.T) {
 	}
 }
 
+func TestOSXKeychainHelperStoreWithUncleanPath(t *testing.T) {
+	helper := Osxkeychain{}
+	creds := &credentials.Credentials{
+		ServerURL: "https://::1:8080//////location/../../hello",
+		Username:  "testuser",
+		Secret:    "testsecret",
+	}
+
+	// Clean store before and after the test.
+	defer helper.Delete(creds.ServerURL)
+	if err := helper.Delete(creds.ServerURL); err != nil && !credentials.IsErrCredentialsNotFound(err) {
+		t.Errorf("prepare: failed to delete '%s': %v", creds.ServerURL, err)
+	}
+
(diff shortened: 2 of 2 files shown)


crazy-max (approved):

osxkeychain/osxkeychain.go
```diff
@@ -121,10 +122,20 @@ func (h Osxkeychain) List() (map[string]string, error) {
...
-		if r.Path == "" {
-			continue
+		proto := "http"
+		if r.Protocol == kSecProtocolTypeHTTPS {
+			proto = "https"
 		}
-		resp[r.Path] = r.Account
+		host := r.Server
+		if r.Port != 0 {
+			host += ":" + strconv.Itoa(int(r.Port))
```
thaJeztah, line 131: Should we used `net.JoinHostPort` here, or was this on purpose to keep the old format?

If that's the case, perhaps we should have a comment here to describe that we did so on purpose.

osxkeychain/osxkeychain.go
```diff
@@ -121,10 +122,20 @@ func (h Osxkeychain) List() (map[string]string, error) {
...
+		}
+		u := url.URL{
+			Scheme: proto,
+			Host:   host,
+			Path:   r.Path,
```
thaJeztah, line 136: Do we know if `Path` is always cleaned? Mostly because `URL.String()` doesn't do any cleaning up; https://go.dev/play/p/azv5LBXYlZo

```go
package main

import (
	"fmt"
	"net/url"
)

func main() {
	u := url.URL{
		Scheme: "https",
		Host:   "::1:8080",
		Path:   "//////location/../../hello",
	}

	fmt.Println(u.String())
}
```

Will print `https://::1:8080//////location/../../hello`

osxkeychain/osxkeychain.go
akerouanton: Nope, that's an oversight. I'm not trying to preserve the old behavior because it seems buggy to me:

- You store a credential for `https://registry.example.com:5000/v1/`
- And you end up with `https://registry.example.com/v1/:5000` in a subsequent `list`

Unless someone starts complaining because they're depending on that specific bug, I don't think we should preserve the old behavior. If that's the case, and we want to add an escape hatch, I think we can introduce an env var. Downside: integrations need to use `client.NewShellProgramFunc` to inherit the calling process' env vars, or `NewShellProgramFuncWithEnv` with that escape hatch specified.

osxkeychain/osxkeychain.go
akerouanton: > Do we know if Path is always cleaned?

It's not, and never was. (I wrote a test checking if uncleaned paths are preserved, and tried to run it on this branch and on v0.8.2 -- it passes on both.)

I'm not sure if that's correct tbh, but at least this PR doesn't introduce yet another backward-incompatible change, so it should be okay from that standpoint.

osxkeychain/osxkeychain.go
thaJeztah: > And you end up with `https://registry.example.com/v1/:5000` in a subsequent list

And now looking at that example, I suspect some of that is related to the assumption that we store creds _per host_, but I suspect the handling of that lived (or still lives) elsewhere.

I.e., the credentials helper would not show anything other than `registry.example.com` or `registry.example.com:5000`, and where needed for the credentials-store used, make that into a proper URL (with scheme); `https://registry.example.com:5000`.

That obviously fails if data was _not_ normalised? And possibly part of that may have been because the "special case" for `https://index.docker.io/v1/`

Also see;

- https://github.com/docker/docker-credential-helpers/issues/256
- https://github.com/docker/cli/issues/190

thaJeztah (approved):
LGTM, thanks!

PR #365, opened 2025-03-03
[DNM] tests: backward-compatibility with v0.8.2

akerouanton:
**- What I did**

We recently introduced bad-looking regressions in v0.9.0. These were not caught by our test suite. This commit adds a new `backward_compatibility` suite to make sure it doesn't happen again.

**- How to verify it**

New GHA steps should be green.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,10 +50,15 @@ jobs:
           - macOS-14
           - macOS-13
           - windows-2022
+    env:
+      TEST_BC_BASE_VERSION: 0.8.2
     steps:
       -
         name: Checkout
         uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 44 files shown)


akerouanton:
Tests are timing out on macOS due to the keychain being locked, and macOS displaying the GUI popup that requests password. I can't find a proper way to bypass that -- maybe `security authorizationdb write …` could do the job but the list of keys it accepts are undocumented and are subject to change across (major) macOS versions.

In retrospect, having tests that ensure backward compatibility is no different than a proper suite of conformance tests, plus human review ensuring tests aren't inadequately changed across versions. Of course, this means we can't ensure backward compatibility between version N (eg. v0.9.2) and version N-1 (eg. v0.8.2), but going forward it should be okay.

So, I'll abandon this PR and work on a shared suite of conformance tests instead.

PR #367, opened 2025-03-14
osxkeychain: store: add atyp attribute

akerouanton:
**- What I did**

Prior to v0.9.0, the osxkeychain creds helper was adding the `atyp` attribute (ie. authentication type) to its credentials. It was also specifying this attribute when querying the keychain for credentials.

Since v0.9.0, we don't set this attribute anymore. So, if a credential is stored with v0.9.0+ and then queried with a v0.8.2 helper, the atyp attribute will be missing and the credential won't be found.

**- How to verify it**

```
- osxkeychain: fix a bug that was preventing credentials created with v0.9.0+ to be retrieved with older versions
```

### DIFF
--- a/osxkeychain/osxkeychain.go
+++ b/osxkeychain/osxkeychain.go
@@ -46,6 +46,17 @@ func (h Osxkeychain) Add(creds *credentials.Credentials) error {
 	item.SetLabel(credentials.CredsLabel)
 	item.SetAccount(creds.Username)
 	item.SetData([]byte(creds.Secret))
+	// Prior to v0.9, the credential helper was searching for credentials with
+	// the "dflt" authentication type (see [1]). Since v0.9.0, Get doesn't use
+	// that attribute anymore, and v0.9.0 - v0.9.2 were not setting it here
+	// either.
+	//
+	// In order to keep compatibility with older versions, we need to store
+	// credentials with this attribute set. This way, credentials stored with
+	// newer versions can be retrieved by older versions.
+	//
+	// [1]: https://github.com/docker/docker-credential-helpers/blob/v0.8.2/osxkeychain/osxkeychain.c#L66
+	item.SetAuthenticationType("dflt")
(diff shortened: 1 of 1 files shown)


osxkeychain/osxkeychain.go
```diff
@@ -46,6 +46,13 @@ func (h Osxkeychain) Add(creds *credentials.Credentials) error {
 	item.SetLabel(credentials.CredsLabel)
 	item.SetAccount(creds.Username)
 	item.SetData([]byte(creds.Secret))
+	// Prior to v0.9.0, the credential helper was searching for credentials with
+	// the "dflt" authentication type. Since this version, Get doesn't use that
+	// attribute anymore.
+	//
+	// In order to keep compatibility with older versions, we need to store
+	// credentials with this attribute set.
+	item.SetAuthenticationType("dflt")
```
akerouanton, line 55: As a follow-up, it'd be great to add this constant and the ones I hardcoded in https://github.com/docker/docker-credential-helpers/pull/361 to `keybase/go-chain`.

I started doing that, but it's lower priority than getting this fixed here before we release the next version of Docker Desktop with a creds store that doesn't support downgrades.

crazy-max: Hum yes indeed I looked at previous implem https://github.com/docker/docker-credential-helpers/pull/282/files#diff-13e5d0d1bee45d43439783edc43138e1d9584d2ec29cab4969460384db72ec42L28 and we set it to `dflt` as well.

I thought `dflt` was the default SecAuthenticationType in `keybase/go-keychain` so agree to set this upstream.

crazy-max (approved):

osxkeychain/osxkeychain.go
akerouanton: Oh! I forgot to add a link to `keychain_get` in my comment. Let me update it before I merge.

PR #375, opened 2025-08-08
update to go1.23.12

austinvazquez:
**- What I did**
update to go1.23.12


thaJeztah (approved):
LGTM

we can probably start updating to go1.24 as well after this

PR #383, opened 2025-10-01
update to go1.24.7, xx v1.7.0

thaJeztah:
- [x] depends on  / stacked on https://github.com/docker/docker-credential-helpers/pull/385

**- What I did**

**- How I did it**

**- How to verify it**

**- Description for the changelog**

**- A picture of a cute animal (not mandatory but encouraged)**

PR #385, opened 2025-10-01
deb: Dockerfile: update to debian bookworm, ubuntu jammy (22.04)

thaJeztah:
- relates to https://github.com/docker/docker-credential-helpers/pull/383

**- What I did**

**- How I did it**

**- How to verify it**


thaJeztah:
Yeah, we should look at that; but definitely didn't want to do so in a `.patch` release. Given that the GitHub repo is currently the only place we publish packages (we don't publish them on `download.docker.com` currently), I also didn't want to just remove them.

PR #412, opened 2026-03-05
Dockerfile: document build-args

thaJeztah:
```bash
docker buildx build --quiet --call=outline .

BUILD ARG            VALUE                                         DESCRIPTION
GO_VERSION           1.25.7                                        sets the version of the golang base image to use.
BASE_DEBIAN_DISTRO   bookworm                                      sets the golang base image debian variant to use.
XX_VERSION           1.7.0                                         sets the version of the xx utility to use.
OSXCROSS_VERSION     11.3-r8-debian                                sets the MacOSX cross toolchain to use.
PACKAGE              github.com/docker/docker-credential-helpers   sets the package name to print in the "--version" output.
```

**- What I did**

**- How I did it**

**- How to verify it**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,28 @@
 # syntax=docker/dockerfile:1
 
+# GO_VERSION sets the version of the golang base image to use.
+# It must be a valid tag in the docker.io/library/golang image repository.
 ARG GO_VERSION=1.25.8
-ARG DEBIAN_VERSION=bookworm
 
+# BASE_DEBIAN_DISTRO sets the golang base image debian variant to use.
+# It must be a valid variant in the docker.io/library/golang image repository.
+ARG BASE_DEBIAN_DISTRO=bookworm
+
+# XX_VERSION sets the version of the xx utility to use.
+# It must be a valid tag in the docker.io/tonistiigi/xx image repository.
 ARG XX_VERSION=1.7.0
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,12 +1,28 @@
...
 ARG OSXCROSS_VERSION=11.3-r8-debian
+
+# GOLANGCI_LINT_VERSION sets the version of the golangci-lint image to use.
+# It must be a valid tag in the docker.io/golangci/golangci-lint image repository.
 ARG GOLANGCI_LINT_VERSION=v2.8
 
+# PACKAGE sets the package name to print in the "--version" output.
+# It sets the "github.com/docker/docker-credential-helpers/credentials.Package
+# variable at compile time.
 ARG PACKAGE=github.com/docker/docker-credential-helpers
```
thaJeztah, line 26: FWIW; it's really unfortunate that it doesn't allow an empty line between; i.e., this doesn't work (doesn't show up in outline);


```dockerfile
# PACKAGE sets the package name to print in the "--version" output.
#
# It sets the "github.com/docker/docker-credential-helpers/credentials.Package
# variable at compile time.
ARG PACKAGE=github.com/docker/docker-credential-helpers
```


Wondering if we should allow that; would that be possible? cc @crazy-max

thaJeztah:
@vvoland @crazy-max PTAL

vvoland (approved):

PR #427, opened 2026-04-20
fix(wincred): Use UTF-16 encoding for passwords (carry 335)

thaJeztah:
- rebased, closes https://github.com/docker/docker-credential-helpers/pull/335

This change adds UTF-16 encoding to the Windows credential helper.

As recommended by the author of the wincred library, the password should be encoded to allow other applications to use the same password.

This is especially needed when the docker image repository is hosted on a JFrog Artifactory, that also hosts Python repositories. Because nobody wants to store their credentials in as plain test within the _pip.ini_, they're going to enable the keyring support for pip, which in turn uses the Win32API that assumes that the password is a UTF-16 encoded string. Since this is not the case pip will crash with an encoding error.

**- What I did**

**- How I did it**

**- How to verify it**


wincred/wincred.go
```diff
@@ -20,10 +25,19 @@ func (h Wincred) Add(creds *credentials.Credentials) error {
...
 	g.Persist = winc.PersistLocalMachine
 	g.Attributes = []winc.CredentialAttribute{{Keyword: "label", Value: credsLabels}}
 
+
+	encoder := unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM).NewEncoder()
+	blob, _, err := transform.Bytes(encoder, []byte(creds.Secret))
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
```
thaJeztah, line 37: changed this to an error return instead

PR #432, opened 2026-05-08
wincred: avoid repeated string->[]byte conversion

thaJeztah:
- taken from https://github.com/docker/docker-credential-helpers/pull/427

### DIFF
--- a/wincred/wincred.go
+++ b/wincred/wincred.go
@@ -11,17 +11,18 @@ import (
 	"github.com/docker/docker-credential-helpers/registryurl"
 )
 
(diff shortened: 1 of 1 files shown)


thaJeztah:
Actually, not sure if this was a good idea; I'm pretty sure Go is able to optimise this; perhaps only not for the Add case, but that's not a massive "hot path", so maybe we should not do this; let me do a quick check.

thaJeztah:
Yeah, confirmed; compiler inlines all those, except for the `Add`, which probably is too little to optimise; opened https://github.com/docker/docker-credential-helpers/pull/433 instead

